PR #4, opened 2021-03-12
Add machine to CI step

fbertsch:


.circleci/config.yml
```diff
@@ -2,6 +2,8 @@
 version: 2
 jobs:
   build:
```
relud, line 4:
```suggestion
  deploy:
```

.circleci/config.yml
```diff
@@ -2,6 +2,8 @@
...
+    machine:
+      image: ubuntu-2004:202101-01
```
relud, line 6: unlike `machine` it works in the circleci cli
```suggestion
    docker:
      - image: cimg/base:2020.01
```

relud (approved):
r+wc

PR #5, opened 2021-03-15
Add on-the-fly modeling to KPI dash

fbertsch:
This does a lot.
1. Unify the KPI and usage explores into one, with on-the-fly modeling for any filters
2. Move the explore to a KPI model, which we can make available to everyone
3. Make measures in the usage explore filters. This means we can't group by those fields, e.g. cannot easily show ten country's CDOU (because we can't pivot on a filter). However it allows us to use on-the-fly modeling.

### DIFF
--- a/KPI/kpi.model.lkml
+++ b/KPI/kpi.model.lkml
@@ -1,0 +1,21 @@
+connection: "telemetry"
+
+include: "./dashboards/*.dashboard"
+include: "./views/*.view.lkml"                # include all views in the views/ folder in this project
+
+explore: firefox_desktop_usage_2021 {
+  label: "Firefox Desktop Usage"
+  group_label: "KPIs"
+  from: firefox_desktop_usage_2021
+  join: prediction {
+    view_label: "Forecast"
+    type: left_outer
+    sql_on: ${firefox_desktop_usage_2021.date} = ${prediction.date} ;;
+    relationship: one_to_one
+  }
+}
+
+# For suggestions
+explore: firefox_desktop_usage_fields {
+  hidden: yes
+}
--- a/KPI/views/firefox_desktop_usage_forecast.view.lkml
+++ b/KPI/views/firefox_desktop_usage_forecast.view.lkml
@@ -1,0 +1,301 @@
+view: dau_model {
+  derived_table: {
+    sql_trigger_value: current_timestamp() ;;
+    sql_create:
+    DECLARE current_key STRING DEFAULT ARRAY_TO_STRING(REGEXP_EXTRACT_ALL("""
+      {% condition firefox_desktop_usage_2021.activity_segment %} activity_segment {% endcondition %}
+      {% condition firefox_desktop_usage_2021.campaign %} campaign {% endcondition %}
+      {% condition firefox_desktop_usage_2021.channel %} channel {% endcondition %}
+      {% condition firefox_desktop_usage_2021.content %} content {% endcondition %}
+      {% condition firefox_desktop_usage_2021.country %} country {% endcondition %}
+      {% condition firefox_desktop_usage_2021.distribution_id %} distribution_id {% endcondition %}
+      {% condition firefox_desktop_usage_2021.id_bucket %} id_bucket {% endcondition %}
+      {% condition firefox_desktop_usage_2021.medium %} medium {% endcondition %}
+      {% condition firefox_desktop_usage_2021.os %} os {% endcondition %}
+      {% condition firefox_desktop_usage_2021.source %} source {% endcondition %}
+      {% condition firefox_desktop_usage_2021.attributed %} (attributed) {% endcondition %}
+      -- add source or campaign not null as attributed
+      -- country tier: FR, GB, US, CA, DE
+      """, r"(\(.*\))"), '');
+
+    DECLARE row_count INT64;
+
+    SET row_count = (
+      SELECT COUNT(*)
+      FROM mozdata.analysis.dou_forecasts
+      WHERE key = current_key);
+
+    IF row_count = 0 THEN
+      -- Create model
+      CREATE OR REPLACE MODEL
+      ${SQL_TABLE_NAME} OPTIONS(
+        model_type='ARIMA',
+        time_series_data_col='dau',
+        time_series_timestamp_col='submission_date'
+      ) AS
+      SELECT
+        submission_date,
+        dau
+      FROM
+        ${firefox_desktop_usage_2021.SQL_TABLE_NAME}
+      WHERE
+        submission_date < '2021-01-19';
+    END IF; ;;
+  }
+}
+
+view: new_profiles_model {
+  derived_table: {
+    sql_trigger_value: current_timestamp() ;;
+    sql_create:
+    DECLARE current_key STRING DEFAULT ARRAY_TO_STRING(REGEXP_EXTRACT_ALL("""
+      {% condition firefox_desktop_usage_2021.activity_segment %} activity_segment {% endcondition %}
+      {% condition firefox_desktop_usage_2021.campaign %} campaign {% endcondition %}
+      {% condition firefox_desktop_usage_2021.channel %} channel {% endcondition %}
+      {% condition firefox_desktop_usage_2021.content %} content {% endcondition %}
+      {% condition firefox_desktop_usage_2021.country %} country {% endcondition %}
+      {% condition firefox_desktop_usage_2021.distribution_id %} distribution_id {% endcondition %}
+      {% condition firefox_desktop_usage_2021.id_bucket %} id_bucket {% endcondition %}
+      {% condition firefox_desktop_usage_2021.medium %} medium {% endcondition %}
+      {% condition firefox_desktop_usage_2021.os %} os {% endcondition %}
+      {% condition firefox_desktop_usage_2021.source %} source {% endcondition %}
+      {% condition firefox_desktop_usage_2021.attributed %} (attributed) {% endcondition %}
+      """, r"(\(.*\))"), '');
+
+    DECLARE row_count INT64;
+
+    SET row_count = (
+      SELECT COUNT(*)
+      FROM mozdata.analysis.dou_forecasts
+      WHERE key = current_key);
+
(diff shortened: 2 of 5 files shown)


KPI/kpi.model.lkml
```diff
@@ -0,0 +1,21 @@
...
+  join: prediction {
+    view_label: "Forecast"
+    type: left_outer
+    sql_on: ${firefox_desktop_usage_2021.date} = ${prediction.date} ;;
+    relationship: one_to_one
+  }
+}
+
+# For suggestions
+explore: firefox_desktop_usage_fields {
```
fbertsch, line 19: Because we have a sql derived_table query for `firefox_desktop_usage_2021` (see e.g. `select submission_date, sum(dau) as dau, ...`), we don't get all the dimension values from that query, so Looker can't use it for the filter suggestions.

Instead we create a parallel, hidden explore that we use for suggestions on those filter fields.

KPI/views/firefox_desktop_usage_forecast.view.lkml
```diff
@@ -0,0 +1,301 @@
...
+
+          CAST(new_profiles AS FLOAT64) AS new_profiles_forecast,
+          CAST(NULL AS FLOAT64) AS new_profiles_target,
+          CAST(NULL AS FLOAT64) AS new_profiles_forecast_lower,
+          CAST(NULL AS FLOAT64) AS new_profiles_forecast_upper,
+        FROM
+          ${firefox_desktop_usage_2021.SQL_TABLE_NAME}
+        WHERE
+          submission_date BETWEEN '2021-01-01' AND '2021-01-18'
+      ), predicted_data AS (
```
fbertsch, line 135: Okay, something happened here. Basically the BQ ML API only gives results for _after_ the last date of your training set. So in this case our training set goes to `2021-01-18`, which caused our forecasted CDOU to be artificially low, since dou and cdou for 01/01-01/18 are forecasted 0.

I deal with this by putting the actual dau in for those dates, which gives us appropriate values for cdou 2021 (exact same up to 2021-01-18). However the plots are odd with a straight line at the beginning.

KPI/views/firefox_desktop_usage_forecast.view.lkml
```diff
@@ -0,0 +1,301 @@
...
+          USING (forecast_timestamp)
+      )
+      SELECT
+        current_key AS key,
+        submission_date,
+
+        dau_forecast,
+        dau_target,
+        SUM(dau_forecast) OVER (ORDER BY submission_date ASC) AS cdou_forecast,
+        (SUM(dau_forecast) OVER (ORDER BY submission_date ASC)) * 1.05 AS cdou_target,
```
irrationalagent, line 165: I wonder if we should remove the targets here in order to avoid confusion. The +5% target is for the overall timeseries and is based on the "official" forecast only. We don't technically have a target for individual DAU segments.

irrationalagent: Also this made me realize that I have misunderstood how this was working when we talked today. Is there any way for us to display the "official" forecasts & target for the top-level unfiltered timeseries, and show the arima models otherwise? These are the targets forecasts that everyone has signed off on and have been tracking in the datastudio dashboard, and we should stick to them when displaying the overall DAU and new profiles time series. Or maybe I'm confused about what's going on here.

fbertsch: Don't worry, it is confusing. We are using the official forecasts / targets for the top-level KPIs. There are a table and associated view, the table has the cache (with the on-the-fly modeling) and the view unions that with the official targets:

```
CREATE OR REPLACE TABLE
      mozdata.analysis.dou_forecasts_cache
CLUSTER BY key
    AS
      SELECT
        CAST(NULL AS STRING) AS key,
        CAST(NULL AS TIMESTAMP) AS submission_date,
        
        -- DAU
        CAST(NULL AS FLOAT64) AS dau_forecast,
        CAST(NULL AS FLOAT64) * 1.05 AS dau_target,
        CAST(NULL AS FLOAT64) AS cdou_forecast,
        CAST(NULL AS FLOAT64) AS cdou_target,
        CAST(NULL AS FLOAT64) AS dau_forecast_lower,
        CAST(NULL AS FLOAT64) AS dau_forecast_upper,
        
        -- New Profiles
        CAST(NULL AS FLOAT64) AS new_profiles_forecast,
        CAST(NULL AS FLOAT64) * 1.05 AS new_profiles_target,
        CAST(NULL AS FLOAT64) AS cum_new_profiles_forecast,
        CAST(NULL AS FLOAT64) AS cum_new_profiles_target,
        CAST(NULL AS FLOAT64) AS new_profiles_forecast_lower,
        CAST(NULL AS FLOAT64) AS new_profiles_forecast_upper,
      FROM
        mozdata.telemetry.clients_daily
      WHERE
        FALSE;

CREATE OR REPLACE VIEW
mozdata.analysis.dou_forecasts
AS
SELECT *
FROM mozdata.analysis.dou_forecasts_cache
UNION ALL
SELECT
  '' AS key,
  CAST(ds AS TIMESTAMP) AS submission_date,
  
  -- DAU
  dau_forecast.dau_forecast,
  dau_forecast.dau_forecast * 1.05 AS dau_target,
  SUM(dau_forecast.dau_forecast) OVER (PARTITION BY date_trunc(ds, YEAR) ORDER BY ds ASC) AS cdou_forecast,
  SUM(dau_forecast.dau_forecast) OVER (PARTITION BY date_trunc(ds, YEAR) ORDER BY ds ASC) * 1.05 AS cdou_target,
  dau_forecast.dau_forecast_lower,
  dau_forecast.dau_forecast_upper,
  
  -- New Profiles
  np_forecast.yhat AS new_profiles_forecast,
  np_forecast.yhat * 1.05 AS new_profiles_target,
  SUM(np_forecast.yhat) OVER (PARTITION BY date_trunc(ds, YEAR) ORDER BY ds ASC) AS cum_new_profiles_forecast,
  SUM(np_forecast.yhat) OVER (PARTITION BY date_trunc(ds, YEAR) ORDER BY ds ASC) * 1.05 AS cum_new_profiles_target,
  np_forecast.yhat_lower AS new_profiles_forecast_lower,
  np_forecast.yhat_upper AS new_profiles_forecast_upper,
FROM
  `mozdata.analysis.loines_desktop_dau_forecast_2021-01-19` AS dau_forecast
JOIN
  `mozdata.analysis.loines_desktop_new_profiles_forecast_2021-01-19` AS np_forecast
  ON dau_forecast.ds = DATE(np_forecast.date);
```

fbertsch: I need to submit the above to `bigquery_etl` so it is centrally defined.

irrationalagent:
OK after reading this a few times and learning about templated filters I think I get what's going on here. I say we merge this for now and play around with it for awhile to see if its useful.

I have a few high level thoughts.

1. Some of the forecasts are going to be really wacky, with the wackiness roughly proportional to how deep people drill (and how far we are out from the training data cutoff). We should probably put a disclaimer on the dashboard saying that only the top-level forecasts have been vetted, and that we only have actual targets/goals for those. If we don't make this clear I know we will be getting a lot of confused questions (we already get a lot for the official forecasts as it is).
2. We've talked about adding up-to-date forecasts, which tell a different story (i.e. where we seem to be headed right now, rather than how we are doing vs. "beginning of year expectations"). Obviously it would be awkward to have both forecast types available on the same dashboard(s). Maybe we should canvass a bit to figure out what people would find most useful. Or, make the up-tod-date ones available elsewhere. (Actually, we could probably configure things so the dash has an option - "Show Forecast: (As of 20219-01-19 | As of Yesterday) )
3. I think there may be a limited set of slices that we would want cross-validated (vetted) prophet forecasts for. Mac/Windows/Linux I know is one cut that people have asked for. Thinking out loud, if we had "official" forecasts for those 3 * (top 10 countries by DAU) * (attributed / unattributed) that would be 60 forecasts, which I think is reasonable if we were to re-run each day. Its still on my list to refactor the prophet forecasting repo. Maybe I can get to that this week. If I do, we would have 2 classes of forecast - official prophet forecasts and auto-made ARIMA ones.

Thanks for doing this!

irrationalagent (approved):

fbertsch:
>     1. Some of the forecasts are going to be really wacky, with the wackiness roughly proportional to how deep people drill (and how far we are out from the training data cutoff). We should probably put a disclaimer on the dashboard saying that only the top-level forecasts have been vetted, and that we only have actual targets/goals for those. If we don't make this clear I know we will be getting a lot of confused questions (we already get a lot for the official forecasts as it is).

Agreed. We could actually display some sort of warning for small populations, basically a text that only shows if the `MAX(recent_dau)` is small enough.

>     2. We've talked about adding up-to-date forecasts, which tell a different story (i.e. where we seem to be headed right now, rather than how we are doing vs. "beginning of year expectations"). Obviously it would be awkward to have both forecast types available on the same dashboard(s). Maybe we should canvass a bit to figure out what people would find most useful. Or, make the up-tod-date ones available elsewhere. (Actually, we could probably configure things so the dash has an option - "Show Forecast: (As of 20219-01-19 | As of Yesterday) )

We could integrate on-the-fly up-to-date forecasts with the existing plots as well. Basically it would show actual DAU for up to `current_date`, and then forecast for the rest of the year. We would have 4 series:
- Target (Over all of 2021)
- 2020 Forecast (Over all of 2021)
- Up-to-date Forecast (Past current date)
- Actual (before current date)

>     3. I think there may be a limited set of slices that we would want cross-validated (vetted) prophet forecasts for. Mac/Windows/Linux I know is one cut that people have asked for. Thinking out loud, if we had "official" forecasts for those 3 * (top 10 countries by DAU) * (attributed / unattributed) that would be 60 forecasts, which I think is reasonable if we were to re-run each day. Its still on my list to refactor the prophet forecasting repo. Maybe I can get to that this week. If I do, we would have 2 classes of forecast - official prophet forecasts and auto-made ARIMA ones.

Right, we can indicate on the page which ones are "official" as well (if all of the simple-prophet ones will be official).

PR #7, opened 2021-03-18
Changes to desktop KPI dash

irrationalagent:
Most of this is changes to the actual dashboard, adding info text, renaming things, etc.

I did make one noticeable change to the KPI views. I joined the `firefox_desktop_usage_2021` onto `moz-fx-data-shared-prod.static.country_codes_v1` to get the readable names of countries, so those will be displayed in the filter instead of the two letter country code. I had to move some things around but it seemed to work when I tested it. We could also consider adding the country names directly to the underlying `firefox_desktop_usage_2021` BQ table, which would make the in-looker join unnecessary. Notice that I coalesce the country code with the country name, so if no matching name is found then it displays the code. this does seem to happen in a couple of cases.

### DIFF
--- a/KPI/dashboards/desktop_kpis_with_filters.dashboard.lookml
+++ b/KPI/dashboards/desktop_kpis_with_filters.dashboard.lookml
@@ -1,12 +1,11 @@
-- dashboard: desktop_kpi_dashboard_with_filters__editable
-  title: Desktop KPI Dashboard with Filters - editable
+- dashboard: desktop_kpi_dashboard_with_filters
+  title: Desktop KPI Dashboard with Filters
   layout: newspaper
   preferred_viewer: dashboards-next
-  load_configuration: wait
-  refresh: 2147484 seconds
+  refresh: 1 day
   elements:
-  - title: Desktop DAU
-    name: Desktop DAU
+  - title: Desktop Daily Active Users (DAU)
+    name: Desktop Daily Active Users (DAU)
     model: kpi
     explore: firefox_desktop_usage_2021
     type: looker_line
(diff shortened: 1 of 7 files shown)


jklukas:
> We could also consider adding the country names directly to the underlying firefox_desktop_usage_2021 BQ table

I am generally in favor of doing the join for country_name at the view level in BQ :+1:

> Notice that I coalesce the country code with the country name, so if no matching name is found then it displays the code. this does seem to happen in a couple of cases.

Can I get to this in Looker? I'm interested to see what these edge cases are.

jklukas:
I looked into country codes, and the only outliers I found were XK and `??`. We inject `??` in clients_daily where MaxMind returns null. See https://github.com/mozilla/bigquery-etl/pull/1902 for adding XK.

jklukas:
Adding country_name to the views in https://github.com/mozilla/bigquery-etl/pull/1903

irrationalagent:
Thanks for jumping in on that jeff! I'll pull down the changes and remove the looker-based join.

fbertsch (changes requested):
We should add a few other updates:
1. All of the top-line number descriptions should be completely present. See attached screenshot, where the long ones have elipses at the end. This is at 100% zoom level, full-screen Firefox, on a 13-inch monitor. On larger monitors/zoom-outs they are present.

Screen Shot 2021-03-18 at 10 43 44 AM

2. Eric recommended switching the "Difference from ..." plots to be relative changes. That will help with the slices we've added, because they have different baseline amounts (e.g. a 100k decrease is much different for release and beta). What do you think?

3. For the documentation, Jennifer [gave me a few possibilities](https://mozilla.cloud.looker.com/dashboards-next/8) if we use HTML rather than markdown. I kind of like having that short section with links to a document.

irrationalagent:
>     1. Eric recommended switching the "Difference from ..." plots to be relative changes. That will help with the slices we've added, because they have different baseline amounts (e.g. a 100k decrease is much different for release and beta). What do you think?

Let's hold off on this change for now. This chart is apparently something that leadership is used to seeing in absolute terms (they've asked about it) so I don't want to change it to percent until we run that by them. Trying to keep the changes from the DS dashboard minimal at this point to avoid confusion. 

I'll see what I can do for the other things.

irrationalagent:
Can you see how this looks on your screen now @fbertsch ? seems to work for me but these things can be weird. 

After we merge this I will have a PR ready for getting the mobile KPIs/on the fly forecasting working for mobile similar to desktop.

fbertsch (approved):
r+wc - just increase the height of those two summary boxes by a hair so there's no vertical scroll bar

PR #9, opened 2021-03-24
Add model for Mozilla VPN

relud:
Trying to develop some views and explores for vpn, but it needs a model. If I understand correctly, adding just the model here should allow me to do that?


relud:
learned that i'm supposed to do this in dev

PR #11, opened 2021-03-31
Add view for legacy event counts via "unified metrics"

acmiyaguchi:
[jira ticket](https://jira.mozilla.com/browse/LOOK-1)

See https://github.com/mozilla/bigquery-etl/pull/1815

This adds a new view for looking at certain event counts in the legacy ios data. The legacy data is backfilled until the beginning of 2020, so going further beyond that may not be useful. It should be possible to backfill the data all the way to the beginning of recorded history if this proves to be useful in some capacity.


firefox_ios/views/unified_metrics.view.lkml
```diff
@@ -0,0 +1,738 @@
...
+
+  dimension: metrics__counter__glean_validation_foreground_count {
+    hidden: yes
+    type: number
+    sql: ${TABLE}.metrics.counter.glean_validation_foreground_count ;;
+    group_label: "Metrics Counter"
+    group_item_label: "Glean Validation Foreground Count"
+  }
+
+  measure: glean_validation_foreground_count {
```
fbertsch, line 381: Some of these have a `count` suffix and some don't. We should ensure they all have the `count` suffix so it's obvious what is being plotted.

firefox_ios/views/unified_metrics.view.lkml
```diff
@@ -0,0 +1,738 @@
...
+    type: string
+    sql: ${TABLE}.key ;;
+  }
+
+  dimension: value {
+    type: number
+    sql: ${TABLE}.value ;;
+  }
+
+  measure: counts {
```
fbertsch, line 596:
```suggestion
  measure: count {
```

Across all these views.

firefox_ios/views/unified_metrics.view.lkml
```diff
@@ -0,0 +1,738 @@
...
+  }
+}
+
+view: unified_metrics__metrics__labeled_counter__tabs_close {
+  dimension: key {
+    type: string
+    sql: ${TABLE}.key ;;
+  }
+
+  dimension: value {
```
fbertsch, line 608: I don't think we should expose `value`
```suggestion
  dimension: value {
    hidden: yes
```

firefox_ios/views/unified_metrics.view.lkml
```diff
@@ -0,0 +1,738 @@
+# Un-hide and use this explore, or copy the joins into another explore, to get all the fully nested relationships from this view
+explore: unified_metrics {
+  hidden: no
+
+  join: unified_metrics__ping_info__experiments {
+    view_label: "Unified Metrics: Ping Info Experiments"
```
fbertsch, line 6:
```suggestion
    view_label: "Experiments"
```

firefox_ios/views/unified_metrics.view.lkml
```diff
@@ -0,0 +1,738 @@
...
+    type: string
+    sql: ${TABLE}.metrics.string.search_default_engine ;;
+    group_label: "Metrics String"
+    group_item_label: "Search Default Engine"
+  }
+
+}
+
+view: unified_metrics__ping_info__experiments {
+  dimension: key {
```
fbertsch, line 565:
```suggestion
  dimension: key {
    label: "Name"
```
or is this `id`?

firefox_ios/views/unified_metrics.view.lkml
```diff
@@ -0,0 +1,738 @@
...
+    type: string
+    sql: ${TABLE}.key ;;
+  }
+
+  dimension: value__branch {
+    type: string
+    sql: ${TABLE}.value.branch ;;
+    group_label: "Value"
+    group_item_label: "Branch"
```
fbertsch, line 574:
```suggestion
    label: "Branch"
```

firefox_ios/views/unified_metrics.view.lkml
```diff
@@ -0,0 +1,738 @@
...
+  }
+
+  dimension: value__extra__type {
+    type: string
+    sql: ${TABLE}.value.extra.type ;;
+    group_label: "Value Extra"
+    group_item_label: "Type"
```
fbertsch, line 581:
```suggestion
    label: "Experiment Type"
```

firefox_ios/views/unified_metrics.view.lkml
```diff
@@ -0,0 +1,738 @@
...
+  }
+
+  measure: counts {
+    type:  sum
+    sql:  ${value} ;;
+  }
+}
+
+view: unified_metrics__metrics__labeled_counter__tabs_close {
+  dimension: key {
```
fbertsch, line 603: Unfortunately the filter suggestions don't work for these joined views. Basically looked tries to query the table `unified_metrics__metrics__labeled_counter__tabs_close` directly, which of course doesn't exist. We have two options: disable filters suggestions for these fields, or create another explore and [use that to load the suggestions for all of these](https://docs.looker.com/data-modeling/learning-lookml/filter-suggestions#suggest_dimension_(field)_and_suggest_explore_(field)).

firefox_ios/views/unified_metrics.view.lkml
acmiyaguchi: This is the name of the actual metric, I don't think they should be suffixed with count (this would be glean_validation_foreground_count_count to be consistent with the naming convention).

acmiyaguchi: I do see the inconsistency between difference measures after looking into it a bit closer, I'll do it one way or the other.

firefox_ios/views/unified_metrics.view.lkml
acmiyaguchi: I'm not sure how another explore would solve this problem without having one explore per unnested field. I don't think there's any harm in having having an empty filter suggest here?

fbertsch: It is a sub-par user experience to leave them empty, but I'm okay adding this as a future improvement if it's too much to ask right now.

firefox_ios/views/unified_metrics.view.lkml
```diff
@@ -0,0 +1,755 @@
...
+
+  measure: count {
+    type:  sum
+    sql:  ${value} ;;
+  }
+}
+
+view: unified_metrics__metrics__labeled_counter__bookmarks_add {
+  dimension: key {
+    label: "Name"
```
fbertsch, line 624: I think I called them "Name", it should be "Label"

fbertsch (approved):
r+wc with a few overall points:
1. Because this is doing LEFT JOIN UNNEST, every key ends up with a `NULL` value which is always 0 and shows up on every plot. We should either figure out a way to remove that null or use CROSS JOIN.
2. We should enable client counting in some fashion. One option is to include a `$COUNTER_NAME_client_count` measure for every counter, and same for the labeled counters, with `type: count distinct` and `sql: client_id`. This will add a number of measures. An alternative is to expose the values again (sorry!), and require that the user filter on them (e.g. a filter on `tab_open_count > 0`), and then just one overall `client_count` field. I'm leaning towards the former, as that's simpler to implement and easier for end users to understand, even if it clutters the explore.

If you want to leave 2.) for follow-up, that's fine. I just know it will come up as a question very soon.

firefox_ios/views/unified_metrics.view.lkml
```diff
@@ -0,0 +1,755 @@
+# Un-hide and use this explore, or copy the joins into another explore, to get all the fully nested relationships from this view
+explore: unified_metrics {
+  hidden: no
+
```
fbertsch, line 4: Ah shoot, one last thing. Generally it's good practice to put the explores in the `model` file, and leave the `view` file with just the view definitions (this is general Looker guidance, but tbh it's sometimes more confusing b/c the understanding the views _requires_ seeing the explore, and vice-versa. So maybe they should just live in one file.)

acmiyaguchi: I'm going to leave this inside the view file for the meanwhile.

acmiyaguchi:
Ooh, I see that I already have approval. This current works, so I'm going to merge this in (which can be updated in a followup if there are bugs found)/

PR #16, opened 2021-04-08
add temporary join to temporary mobile new profiles table.

irrationalagent:
The query for the temp table is [here](https://console.cloud.google.com/bigquery?sq=630180991450:261b6210dc8047da9dfd4e51cbc22efb). It could probably use a look by a DE.

This will allow us to add new profiles to the mobile KPI dash, but we will need to revert/change this back when the clients_first_seen work is totally done. We should also be clear that these numbers could change once that happens.

### DIFF
--- a/KPI/views/mobile_usage_2021.view.lkml
+++ b/KPI/views/mobile_usage_2021.view.lkml
@@ -2,31 +2,49 @@ view: mobile_usage_2021 {
   derived_table: {
     sql:
       with
-        base as (
-      select
-          submission_date,
-          app_name,
-          canonical_app_name,
-          sum(dau) as dau,
-          sum(wau) as wau,
-          sum(mau) as mau,
-          sum(cdou) as cdou
-      from
-          ${mobile_usage_fields.SQL_TABLE_NAME} AS mobile_usage_fields
-      where
-          {% condition mobile_usage_2021.campaign %} campaign {% endcondition %}
-          AND {% condition mobile_usage_2021.channel %} channel {% endcondition %}
-          AND {% condition mobile_usage_2021.country %} country {% endcondition %}
-          AND {% condition mobile_usage_2021.country_name %} country_name {% endcondition %}
-          AND {% condition mobile_usage_2021.distribution_id %} distribution_id {% endcondition %}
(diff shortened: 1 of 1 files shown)


jklukas:
> This will allow us to add new profiles to the mobile KPI dash, but we will need to revert/change this back when the clients_first_seen work is totally done. We should also be clear that these numbers could change once that happens.

Is the urgency here sufficient to warrant this solution? Is `loines_temp_mobile_new_profiles_per_day` going to be updated on a regular basis?

irrationalagent:
> > This will allow us to add new profiles to the mobile KPI dash, but we will need to revert/change this back when the clients_first_seen work is totally done. We should also be clear that these numbers could change once that happens.
> 
> Is the urgency here sufficient to warrant this solution? Is `loines_temp_mobile_new_profiles_per_day` going to be updated on a regular basis?

No I wasn't planning to update it. But it allows me to show stakeholders what the new profile section of the dashboard looks like in the meantime. Do we have an ETA for the cfs work?

jklukas:
> Do we have an ETA for the cfs work?

@acmiyaguchi is actively working on getting cfs integrated to the daily and last_seen tables, and I'm reviewing. I'll let him comment on ETA.

irrationalagent:
OK maybe that's soon enough to hold off on this... we have a meeting with stakeholders next tue and like to deliver if possible a working version of the mobile dashboard at that point.

acmiyaguchi:
Ideally, I'd like to integrate the fields by the end of the week, but it may happen on Monday. I'll be updating https://jira.mozilla.com/browse/DS-1424 appropriately.

fbertsch:
Let's hold off on this until EOD, per Anthony. We can merge it in tomorrow if new_profiles is still not ready from the base tables.

acmiyaguchi:
The `baseline_clients_first_seen` tables are available now (and take into account migration pings in fenix). The `first_seen_date` and `is_new_profile` fields also in the `baseline_clients_daily` and `baseline_clients_last_seen` tables.

irrationalagent:
I'm going to close this since we now have the proper first_seen_dates.

thanks again @acmiyaguchi !

PR #17, opened 2021-04-13
mobile new profiles and kpi dashboard

irrationalagent:
So I didn't end up putting in the new profiles based on first_seen_date yet, for a couple reasons:

1. first_seen_date and new_profiles are not yet incorporated downstream into the 2021_mobile_usage dataset (I filed a bug [here](https://github.com/mozilla/bigquery-etl/issues/1954))
2. I don't think first_seen_date has been added to the the core_clients_last_seen dataset yet, and we need that for focus android new profiles (since its not on glean).

So I ended up just using what we're using right now in the datastudio dash, which is based on profile creation date from `firefox_nondesktop_day_2_7_activation`.

This has a lookml dashboard for the mobile KPIs, so I'm also going to close the other PR on the shared branch for now in the name of expediency.

### DIFF
--- a/KPI/dashboards/mobile_kpis.dashboard.lookml
+++ b/KPI/dashboards/mobile_kpis.dashboard.lookml
@@ -2,7 +2,7 @@
   title: Mobile KPIs
   layout: newspaper
   preferred_viewer: dashboards-next
-  refresh: 12 hours
+  refresh: 2147484 seconds
   elements:
   - title: Firefox for Android (Fennec + Fenix) Daily Active Users (DAU)
     name: Firefox for Android (Fennec + Fenix) Daily Active Users (DAU)
@@ -59,6 +59,7 @@
     defaults_version: 1
     listen:
       Date: mobile_usage_2021.date
+      Country: mobile_usage_2021.country
     row: 17
     col: 12
     width: 12
@@ -118,6 +119,7 @@
     defaults_version: 1
     listen:
       Date: mobile_usage_2021.date
+      Country: mobile_usage_2021.country
     row: 72
     col: 11
     width: 13
@@ -177,6 +179,7 @@
     defaults_version: 1
     listen:
       Date: mobile_usage_2021.date
+      Country: mobile_usage_2021.country
     row: 45
     col: 12
     width: 12
@@ -236,6 +239,7 @@
     defaults_version: 1
     listen:
       Date: mobile_usage_2021.date
+      Country: mobile_usage_2021.country
     row: 98
     col: 11
     width: 13
@@ -261,6 +265,7 @@
     defaults_version: 1
     listen:
       Date: mobile_usage_2021.date
+      Country: mobile_usage_2021.country
     row: 0
     col: 0
     width: 24
@@ -286,6 +291,7 @@
     defaults_version: 1
     listen:
       Date: mobile_usage_2021.date
+      Country: mobile_usage_2021.country
     row: 54
     col: 0
     width: 24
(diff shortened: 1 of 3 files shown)


KPI/dashboards/mobile_kpis.dashboard.lookml
```diff
@@ -0,0 +1,1256 @@
...
+    series_types: {}
+    series_labels:
+      mobile_usage_2021.delta_from_target_count: Difference From Target
+      mobile_usage_2021.delta_from_forecast_count: Difference From Forecast
+    reference_lines: [{reference_type: line, range_start: max, range_end: min, margin_top: deviation,
+        margin_value: mean, margin_bottom: deviation, label_position: center, color: "#000000",
+        line_value: '0', label: At Target (Left) / Forecast (Right)}]
+    defaults_version: 1
+    listen:
+      Date: mobile_usage_2021.date
```
fbertsch, line 698: Not all of the filters are listening to the country filter
```suggestion
      Date: mobile_usage_2021.date
      Country: mobile_usage_2021.country
```

KPI/dashboards/mobile_kpis.dashboard.lookml
```diff
@@ -0,0 +1,1256 @@
...
+    y_axis_combined: true
+    show_null_points: true
+    interpolation: linear
+    y_axes: [{label: '', orientation: left, series: [{axisId: mobile_usage_2021.delta_from_target_count,
+            id: mobile_usage_2021.delta_from_target_count, name: Delta From Target
+              Count}], showLabels: true, showValues: true, unpinAxis: false, tickDensity: default,
+        tickDensityCustom: 5, type: linear}, {label: !!null '', orientation: right,
+        series: [{axisId: mobile_usage_2021.delta_from_forecast_count, id: mobile_usage_2021.delta_from_forecast_count,
+            name: Delta From Forecast Count}], showLabels: true, showValues: true,
+        unpinAxis: false, tickDensity: default, tickDensityCustom: 5, type: linear}]
```
fbertsch, line 744: Just want to see what you think about pinning both to the same axis. The plots may be a bit more understandable that way, but perhaps they didn't look right when you tried it.

KPI/views/mobile_usage_2021.view.lkml
```diff
@@ -2,31 +2,55 @@ view: mobile_usage_2021 {
...
+        from
+            ${mobile_usage_fields.SQL_TABLE_NAME} AS mobile_usage_fields
+        where
+            {% condition mobile_usage_2021.campaign %} campaign {% endcondition %}
+            AND {% condition mobile_usage_2021.channel %} channel {% endcondition %}
+            AND {% condition mobile_usage_2021.country %} country {% endcondition %}
+            AND {% condition mobile_usage_2021.country_name %} country_name {% endcondition %}
+            AND {% condition mobile_usage_2021.distribution_id %} distribution_id {% endcondition %}
+            AND {% condition mobile_usage_2021.id_bucket %} id_bucket {% endcondition %}
+            AND {% condition mobile_usage_2021.os %} os {% endcondition %}
```
fbertsch, line 23: We haven't exposed any of these other filters in the dash, do we want to?

irrationalagent: So right now the table I'm drawing the new profiles from really only has the `country` column that matches the columns from the mobile (DAU) KPI dataset. so filtering on anything other than country would filter the dau/cdou charts but not the new profiles. Once we have the first_seen_date-based new profile counts, we should add them back to the dash.

fbertsch (approved):
LGTM

PR #18, opened 2021-04-13
Add "in_cache" feature to desktop KPI report

fbertsch:
This adds a new element that replaces the previous message (loads in 2-3 seconds after re-running). See below for a quick demo.

https://user-images.githubusercontent.com/20819040/114606323-23241b00-9c69-11eb-915b-023506fa31df.mov

### DIFF
--- a/KPI/views/firefox_desktop_usage_forecast.view.lkml
+++ b/KPI/views/firefox_desktop_usage_forecast.view.lkml
@@ -429,6 +429,6 @@ view: key_in_cache {
 
   measure: is_cached {
     type: string
-    sql: if(MAX(row_count) > 0, "Retrieving cached value...", "Building forecast model, will take 2-3 minutes...") ;;
+    sql: if(MAX(row_count) > 0, "Forecast for current selection is cached.", "Building forecast model, will take 2-3 minutes...") ;;
   }
 }
--- a/KPI/views/firefox_desktop_usage_forecast.view.lkml
+++ b/KPI/views/firefox_desktop_usage_forecast.view.lkml
@@ -405,3 +405,30 @@ view: prediction {
     hidden: yes
   }
 }
+
+view: key_in_cache {
+  derived_table: {
+    sql:
+      SELECT COUNT(*) AS row_count
+      FROM mozdata.analysis.dou_forecasts
+      WHERE key = ARRAY_TO_STRING(REGEXP_EXTRACT_ALL("""
+      {% condition firefox_desktop_usage_2021.activity_segment %} activity_segment {% endcondition %}
+      {% condition firefox_desktop_usage_2021.campaign %} campaign {% endcondition %}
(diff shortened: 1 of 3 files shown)


KPI/views/firefox_desktop_usage_forecast.view.lkml
```diff
@@ -405,3 +405,30 @@ view: prediction {
...
+      {% condition firefox_desktop_usage_2021.medium %} medium {% endcondition %}
+      {% condition firefox_desktop_usage_2021.os %} os {% endcondition %}
+      {% condition firefox_desktop_usage_2021.source %} source {% endcondition %}
+      {% condition firefox_desktop_usage_2021.attributed %} (attributed) {% endcondition %}
+      """, r"(\(.*\))"), '') ;;
+  }
+
+  measure: is_cached {
+    type: string
+    sql: if(MAX(row_count) > 0, "Retreiving cached value...", "Building forecast model, will take 2-3 minutes...") ;;
```
jklukas, line 432:
```suggestion
    sql: if(MAX(row_count) > 0, "Retrieving cached value...", "Building forecast model, will take 2-3 minutes...") ;;
```

I before e, except after c

Or when sounded as 'a' as in 'neighbor' and 'weigh'

Unless the 'c' is part of a 'sh' sound as in 'glacier'

Or it appears in comparatives and superlatives like 'fancier'

And also except when the vowels are sounded as 'e' as in 'seize'

Or 'i' as in 'height'

Or also in '-ing' inflections ending in '-e' as in 'cueing'

Or in compound words as in 'albeit'

Or occasionally in technical words with strong etymological links to their parent languages as in 'cuneiform'

Or in other numerous and random exceptions such as 'science', 'forfeit', and 'weird'.

- [Merriam-Webster](https://www.merriam-webster.com/words-at-play/i-before-e-except-after-c)

jklukas:
This is cool. Does the message persist after the page has finished loading?

fbertsch:
> This is cool. Does the message persist after the page has finished loading?

It does. I didn't experiment with trying to figure out if the dashboard is loaded, I can't think of any Looker capabilities for that :/

irrationalagent (approved):
super clever, thanks for doing this frank it really improves the usability.

fbertsch:
Added a refresh. Now when the cache is filled it will switch to the other text. Should we change it to something like `Slice is cached`? cc @irrationalagent @jklukas

irrationalagent:
> Added a refresh. Now when the cache is filled it will switch to the other text. Should we change it to something like `Slice is cached`? cc @irrationalagent @jklukas

Maybe "The forecast for the current selection has now been cached for future use?" or is that too much verbiage? Trying to think of a less technical term than "slice"

fbertsch:
Okay, I updated it to `The forecast for the current selection is cached.`. That should make it obvious what the difference is.

PR #19, opened 2021-04-14
fix axis problem on desktop lookml dashboard

irrationalagent:
the dau charts on the desktop dash had a hard-coded minimum value which would prevent data from being displayed when a filter was applied.


KPI/dashboards/desktop_kpis_with_filters_alternate.dashboard.lookml
```diff
@@ -601,6 +604,7 @@
     conditional_formatting_include_nulls: false
     series_types: {}
     defaults_version: 1
+    refresh: '10000'
```
fbertsch, line 607: What is this refresh rate? 10k seconds? 10 seconds is nice because it lets you know that it has finished updating

fbertsch (approved):
r+wc

KPI/dashboards/desktop_kpis_with_filters_alternate.dashboard.lookml
irrationalagent: i don't know how that happened maybe it just changed it to 10k ms == 10 seconds??? ill change it back in any case

PR #20, opened 2021-04-14
[Bug 1704515]  Enable a count and fraction of users who set Firefox as `default_browser` in UJET funnel analysis

scholtzan:
For https://bugzilla.mozilla.org/show_bug.cgi?id=1704515

I just want to make sure I'm on the right track here

### DIFF
--- a/user_journey/user_journey.model.lkml
+++ b/user_journey/user_journey.model.lkml
@@ -16,6 +16,15 @@ explore: funnel_analysis {
               OR (${days_of_use.submission_date} = DATE_SUB(current_date, INTERVAL 2 DAY)
                   AND DATE_ADD(${funnel_analysis.submission_date}, INTERVAL 56 DAY) >  DATE_SUB(current_date, INTERVAL 2 DAY)));;
   }
+  join: client_properties {
+    relationship: one_to_one
+    type: left_outer
+    sql_where: client_properties.submission_date BETWEEN DATE_SUB(DATE({% date_start funnel_analysis.date %}), INTERVAL {% parameter client_properties.diff_days %} DAY) AND DATE_SUB(DATE({% date_end funnel_analysis.date %}), INTERVAL {% parameter client_properties.diff_days %} DAY);;
+    sql_on: ${funnel_analysis.sample_id} = ${client_properties.sample_id}
+      AND ${funnel_analysis.client_id} = ${client_properties.client_id}
+      AND ${funnel_analysis.submission_date} = DATE_SUB(${client_properties.submission_date}, INTERVAL {% parameter client_properties.diff_days %} DAY);;
+    fields: [client_properties.fraction_is_default_browser, client_properties.is_default_browser, client_properties.count_is_default_browser, diff_days]
+  }
   join: event_type_1 {
     relationship: many_to_one
     type: cross
@@ -101,14 +110,14 @@ explore: cohort_analysis {
     relationship: many_to_one
     type: cross
   }
-  join: clients_last_seen {
+  join: search_clients_last_seen {
     fields: []
     relationship: one_to_one
     type: left_outer
-    sql_on: ${cohort_analysis.sample_id} = ${clients_last_seen.sample_id}
-        AND ${cohort_analysis.client_id} = ${clients_last_seen.client_id}
-        AND (${cohort_analysis.submission_date} = DATE_SUB(${clients_last_seen.submission_date}, INTERVAL 56 DAY)
-              OR (${clients_last_seen.submission_date} = DATE_SUB(current_date, INTERVAL 2 DAY)
+    sql_on: ${cohort_analysis.sample_id} = ${search_clients_last_seen.sample_id}
+        AND ${cohort_analysis.client_id} = ${search_clients_last_seen.client_id}
+        AND (${cohort_analysis.submission_date} = DATE_SUB(${search_clients_last_seen.submission_date}, INTERVAL 56 DAY)
+              OR (${search_clients_last_seen.submission_date} = DATE_SUB(current_date, INTERVAL 2 DAY)
                   AND DATE_ADD(${cohort_analysis.submission_date}, INTERVAL 56 DAY) >  DATE_SUB(current_date, INTERVAL 2 DAY)));;
   }
   sql_always_where: cohort_analysis.submission_date > "2010-01-01" ;;
@@ -171,7 +180,7 @@ explore: event_counts {
     limit: 500
   }
   sql_always_where: DATE(submission_timestamp) > '2020-01-01'
-  ;;
+    ;;
 }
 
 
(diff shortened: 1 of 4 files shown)


user_journey/views/funnel_analysis/client_properties.view.lkml
```diff
@@ -0,0 +1,58 @@
+view: client_properties {
+  derived_table: {
```
scholtzan, line 2: I also tried simply using the `clients_last_seen` view from looker-hub. However that required to rename the existing `clients_last_seen` view in `cohort_analysis/` and I'm not sure how convenient it is to have all the `clients_last_seen` fields listed in the explore.

fbertsch: We should be able to use [`from`](https://docs.looker.com/reference/explore-params/from-for-join) to specify the Looker view without creating a new one, and [`fields`](https://docs.looker.com/reference/explore-params/fields-for-join) to specify which fields to expose.

Alternatively, [extend](https://docs.looker.com/reference/view-params/extends-for-view) that view with this one, and just add those new measures you need at the bottom.

user_journey/user_journey.model.lkml
```diff
@@ -16,6 +16,14 @@ explore: funnel_analysis {
               OR (${days_of_use.submission_date} = DATE_SUB(current_date, INTERVAL 2 DAY)
                   AND DATE_ADD(${funnel_analysis.submission_date}, INTERVAL 56 DAY) >  DATE_SUB(current_date, INTERVAL 2 DAY)));;
   }
+  join: client_properties {
+    view_label: "Client Properties"
```
fbertsch, line 20: This is unnecessary if it matches the join table name (it does, here).

user_journey/user_journey.model.lkml
```diff
@@ -16,6 +16,14 @@ explore: funnel_analysis {
...
+    relationship: one_to_one
+    type: left_outer
+    sql_on: ${funnel_analysis.sample_id} = ${client_properties.sample_id}
+        AND ${funnel_analysis.client_id} = ${client_properties.client_id}
+        AND ${funnel_analysis.submission_date} = ${client_properties.submission_date};;
```
fbertsch, line 25: This is what's not clear to me. We may want to look at some future date. We should get some insight from Alex about _when_ they are expecting to see this change appear in clients.

user_journey/views/cohort_analysis/search_clients_last_seen.view.lkml
```diff
@@ -1,4 +1,4 @@
-view: clients_last_seen {
+view: search_clients_last_seen {
```
scholtzan, line 1: Extending from `clients_last_seen` requires renaming this view which is also called `clients_last_seen`

fbertsch: Got it, forgot this was `search_clients_last_seen`. Then it doesn't make sense to extend from it (different table)

user_journey/views/funnel_analysis/client_properties.view.lkml
```diff
@@ -0,0 +1,25 @@
+include: "//looker-hub/firefox_desktop/views/clients_last_seen.view.lkml"
+
+view: client_properties {
+  extends: [clients_last_seen]
+
+  parameter: diff_days {
+    type: number
+    default_value: "0"
+  }
```
scholtzan, line 9: Having a parameter that allows to specify the day difference is probably the most flexible option and would allow to comparison with future dates

user_journey/views/cohort_analysis/search_clients_last_seen.view.lkml
scholtzan: I do think it makes sense to extend from the `clients_last_seen` in looker-hub and rename the custom view here to `search_clients_last_seen`, then the naming would also be correct and cause less confusion.

user_journey/user_journey.model.lkml
```diff
@@ -16,6 +16,15 @@ explore: funnel_analysis {
               OR (${days_of_use.submission_date} = DATE_SUB(current_date, INTERVAL 2 DAY)
                   AND DATE_ADD(${funnel_analysis.submission_date}, INTERVAL 56 DAY) >  DATE_SUB(current_date, INTERVAL 2 DAY)));;
   }
+  join: client_properties {
+    relationship: one_to_one
+    type: left_outer
+    sql_where: client_properties.submission_date BETWEEN DATE_SUB(DATE({% date_start funnel_analysis.date %}), INTERVAL {% parameter client_properties.diff_days %} DAY) AND DATE_SUB(DATE({% date_end funnel_analysis.date %}), INTERVAL {% parameter client_properties.diff_days %} DAY);;
```
fbertsch, line 22: Can you confirm that this results in the appropriate partition selection in BigQuery? Sometimes Looker uses timestamp types and BQ isn't able to push down the predicate.

user_journey/views/funnel_analysis/client_properties.view.lkml
```diff
@@ -0,0 +1,25 @@
...
+
+  measure: count_is_default_browser {
+    type: count
+
```
fbertsch, line 13:
```suggestion
```

user_journey/views/funnel_analysis/client_properties.view.lkml
```diff
@@ -0,0 +1,25 @@
...
+view: client_properties {
+  extends: [clients_last_seen]
+
+  parameter: diff_days {
+    type: number
+    default_value: "0"
+  }
+
+  measure: count_is_default_browser {
```
fbertsch, line 11: It'd be good practice for us to start adding descriptions for these important fields.

user_journey/user_journey.model.lkml
```diff
@@ -16,6 +16,15 @@ explore: funnel_analysis {
...
+    sql_on: ${funnel_analysis.sample_id} = ${client_properties.sample_id}
+      AND ${funnel_analysis.client_id} = ${client_properties.client_id}
+      AND ${funnel_analysis.submission_date} = DATE_SUB(${client_properties.submission_date}, INTERVAL {% parameter client_properties.diff_days %} DAY);;
+    fields: [client_properties.fraction_is_default_browser, client_properties.count_is_default_browser, client_properties.is_default_browser, diff_days]
```
fbertsch, line 26: I'm not sure we want to expose `client_properties.is_default_browser`. How do you see that being used in analysis?

scholtzan: I don't think it is useful for analysis. However, removing it from this list results in a LookML Error: 

`Unknown field "client_properties.is_default_browser" referenced in "filters" parameter.
client_properties.view:13 user_journey:funnel_analysis`

user_journey/user_journey.model.lkml
scholtzan: Yes, it does use the right partition. Initially, I was running into that issue but after explicitly casting everything to `DATE`s here it worked correctly.

user_journey/user_journey.model.lkml
fbertsch: Okay, let's include it but hide it then (using `hidden: "yes"`).

fbertsch (approved):
r+wc. Nice job!

PR #21, opened 2021-04-15
Add glean dictionary links for demo

wlach:
Doing this via GitHub, as Looker seems to not want me to commit because of errors.


fbertsch (changes requested):
For us to push this to production, I think we need to enable the links and descriptions for all the metrics, as well as the `*_client_count` metrics.

fbertsch (approved):
Just tested it out, this is great!

PR #22, opened 2021-04-16
Add mozilla vpn dashboard

relud:


mozilla-vpn/explores/funnel.explore.lkml
```diff
@@ -0,0 +1,56 @@
+include: "../views/*.view"
+
+explore: funnel {
+  view_name: login_flows
```
fbertsch, line 4: Consider renaming all of these to event-based names (e.g. begin_funnel, graduated_to_user, graduated_to_strip_customer, etc.)

mozilla-vpn/explores/subscriptions.explore.lkml
```diff
@@ -0,0 +1,267 @@
...
+        STRUCT(
+          ${end_raw} AS date,
+          "Cancelled" AS type,
+          ${cancel_reason} AS granular_type,
+          -1 AS delta
+        ),
+        -- repeat events for displaying net change
+        STRUCT(
+          ${start_raw} AS date,
+          "Net Paid Subscribers" AS type,
```
fbertsch, line 193: IIUC, we don't need these now with the ROW TOTALS

fbertsch (approved):
r+, I definitely missed some stuff we talked about during the walk-through though. By and large this is correct though, and since performance isn't a concern everything else is just semantics.

PR #23, opened 2021-04-19
[Bug 1704544] Enable `is_taskbar_pinned` in Looker UJET funnel explores

scholtzan:
https://bugzilla.mozilla.org/show_bug.cgi?id=1704544

Backfill is currently running for clients_daily and clients_last_seen. So data should be available later today.


user_journey/views/funnel_analysis/client_aggregates.view.lkml
```diff
@@ -1,6 +1,6 @@
 include: "//looker-hub/firefox_desktop/views/clients_last_seen.view.lkml"
 
-view: client_properties {
+view: client_aggregates {
```
scholtzan, line 3: Is `client_aggregates` a better name here?

fbertsch: Hmm, maybe `browser_properties`? I think `client_properties` made more sense than `client_aggregates`.

fbertsch (approved):
LGTM

PR #24, opened 2021-04-20
add total mobile kpis

irrationalagent:
This adds in the total mobile KPIs (all 4 mobile browsers summed together).

Unfortunately this required me to re-do the underlying mobile forecast tables - I created new v2 versions but I'm realizing that I THINK I can just union with the old versions so we don't lose those forecasts. I'll do that tomorrow. 

Also this is on a shared branch so anyone should be able to make commits

edit, [here's the query](
https://console.cloud.google.com/bigquery?sq=630180991450:2a6b960ffd6d400b9123188b8f80d292
) that adds in the total mobile baseline forecast

### DIFF
--- a/KPI/dashboards/mobile_kpis.dashboard.lookml
+++ b/KPI/dashboards/mobile_kpis.dashboard.lookml
@@ -60,7 +60,7 @@
     listen:
       Date: mobile_usage_2021.date
       Country: mobile_usage_2021.country
-    row: 17
+    row: 43
     col: 12
     width: 12
     height: 10
@@ -120,7 +120,7 @@
     listen:
       Date: mobile_usage_2021.date
       Country: mobile_usage_2021.country
-    row: 72
+    row: 98
     col: 11
     width: 13
     height: 9
@@ -180,7 +180,7 @@
     listen:
       Date: mobile_usage_2021.date
       Country: mobile_usage_2021.country
-    row: 45
+    row: 71
     col: 12
     width: 12
     height: 9
@@ -240,7 +240,7 @@
     listen:
       Date: mobile_usage_2021.date
       Country: mobile_usage_2021.country
-    row: 98
+    row: 124
     col: 11
     width: 13
     height: 8
@@ -266,7 +266,7 @@
     listen:
       Date: mobile_usage_2021.date
       Country: mobile_usage_2021.country
-    row: 0
+    row: 26
     col: 0
     width: 24
     height: 8
@@ -292,7 +292,7 @@
     listen:
       Date: mobile_usage_2021.date
       Country: mobile_usage_2021.country
-    row: 54
+    row: 80
     col: 0
     width: 24
     height: 9
@@ -318,7 +318,7 @@
     listen:
       Date: mobile_usage_2021.date
       Country: mobile_usage_2021.country
-    row: 27
+    row: 53
     col: 0
     width: 24
     height: 9
@@ -344,7 +344,7 @@
     listen:
       Date: mobile_usage_2021.date
       Country: mobile_usage_2021.country
-    row: 81
+    row: 107
     col: 0
     width: 24
     height: 8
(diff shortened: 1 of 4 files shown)


irrationalagent:
So maybe I did too much here to avoid starting over on the forecasts. Maybe to avoid the added complexity we should just say screw it and start with a blank slate. But here's what I did:

1. We needed to add the baseline forecasts for the sum of the mobile apps to the cache, so I created mozdata.analysis.mobile_forecasts_official_v2 which contains adds the forecast under the `total_mobile` `app_name` 
2.  I didn't want to overwrite the existing forecasts that we already had cached. So I created mozdata.analysis.mobile_forecasts_v2 and mozdata.analysis.mobile_forecasts_cached_v2  and unioned the existing forecasts (from the "v1" tables) into those tables.
3. To avoid regenerating duplicate forecasts for app_name/country combinations we already had cached, I modified the forecast logic to only generate new forecasts for app_name/country combinations that aren't already in the cache. So for example `US` already had forecasts for all `app_name`s except `total_mobile`, so the new logic will detect that and only create that forecast rather re-forecasting ALL of the app_names again. 

hope that makes sense 😬 if you think we should just nuke the existing mobile forecasts and start clean, that's fine with me.

fbertsch (commented):
@irrationalagent this is quite a bit! I think it might be worth it to take a step back and think about what we're trying to do.

The underlying table has one row for every slice. So right now, we should be able to get the DAU/new_profiles across all products by filtering on the ones we want (all the mobile ones for 2021 KPIs) and then doing a group by on day.

The problem is that the current iteration of the `derived_table` first groups by `app_name` and `canonical_app_name`. I wonder if we can use just `canonical_app_name`, and drop `app_name` entirely from this setup (and ignore it in the underlying table). Then we can also add a filter in the where clause on `canonical_app_name`, e.g.:
```
where
  {% condition mobile_usage_2021.canonical_app_name %} canonical_app_name {% endcondition %}
  ```
  
We would have to change `canonical_app_name` to be a filter rather than a dimension. I checked the dash though and it looks like we only ever use it as a filter anyways.

If we used this approach, we would also lose the capability of forecasting across all apps in one go (the `time_series_id_col`). In exchange we would let people choose subsets of apps, e.g. android-only, or focus (android + ios). Clearly your use case is all mobile apps. If we went this route the cache would not need to be changed.

Thoughts? Let me know if you want to hop on zoom to chat.

irrationalagent:
So finally getting back to it, I think there's an issue with this approach. The targets are defined on a per-app basis, and the forecasting code joins the target lifts to the forecast (and computes the target levels by multiplying the two)  by `app_name`. 

So in the case where a combination of `app_name`s are selected, the target level for *the combination* is not straightforward to compute. We can't just average the targets for the multiple apps together equally; the combined target needs to be weighted toward the app_names that contribute the most to the combined forecast. 

For example, if focus android and fenix are selected we would need to compute the target lift as: 

(proportion of training data that is fenix) * (fenix target_lift) + (proportion of training data that is focus) * (focus target_lift)

Given this, I'm no longer sure what the best path forward is here. Also, I'm concurrently working on an "official" H2 re-forecast for mobile, with the thought that the goals (target lifts) might change. If that's the case, our on-the-fly forecasts here would also become outdated, since the new goals would be based on training data up until the end of H1, and the code currently only forecasts with data starting 2021-01-19. If that happens, we would be starting over with new per-segment forecasts anyway. 

Maybe it makes sense to hold off adding the total mobile forecasts / data until we confirm what the H2 targets are, and whether or not they will be based on forecasts that use data through H1. If they do, then we will have to reset the forecast cache anyway, and that would be the time to add in the total mobile view. 

cc @ethompso28 the last two paragraphs here are relevant to our discussion today. adding total mobile views may not be too straightforward. 

Another option would be to just remove the segmented forecasts for mobile completely.

irrationalagent:
closing in favor of https://github.com/mozilla/looker-spoke-default/pull/53

PR #33, opened 2021-04-26
Bug 1707576 - Add experiment dimensions to UJET explores

fbertsch:
- Adds `experiment_names` explore for suggestions
- Lies about the `one_to_one` relationship between `funnel_analysis` and `experiments`. Deals with this by requiring the user select both of those fields if one of them is filtered on, and requires experiment be selected if branch is selected.
- Does not lie about the `one_to_many` relationship in `event_counts`, because there users can use `client_count` (and it has a primary key to allow for symmetric aggregates)

### DIFF
--- a/user_journey/user_journey.model.lkml
+++ b/user_journey/user_journey.model.lkml
@@ -6,6 +6,25 @@ include: "/user_journey/views/*.view.lkml"
 
 explore: funnel_analysis {
   view_label: " User-Day Funnels"
+  join: experiments {
+    relationship: one_to_one
+    type: cross
+    sql_where:
+      {% if experiments.branch._is_filtered or experiments.experiment._is_filtered %}
+        {% unless experiments.branch._is_selected and experiments.experiment._is_selected %}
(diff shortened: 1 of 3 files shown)


user_journey/user_journey.model.lkml
```diff
@@ -6,6 +6,25 @@ include: "/user_journey/views/*.view.lkml"
 
 explore: funnel_analysis {
   view_label: " User-Day Funnels"
```
scholtzan, line 8:
```suggestion
  view_label: "User-Day Funnels"
```

scholtzan (approved):

user_journey/user_journey.model.lkml
fbertsch: I added this space a dumb hack to have the view appear at the top of the views list, rather than the bottom:
Screen Shot 2021-04-26 at 6 06 36 PM

scholtzan: Ah okay :D

PR #35, opened 2021-04-27

fbertsch:


acmiyaguchi (approved):
This is the same as https://github.com/mozilla/looker-spoke-default/pull/32, but for a different non-prod branch?

fbertsch:
Hah yes, this is for testing on the dev server. Probably should not require a PR for this test mode.

PR #37, opened 2021-04-28
use an array for mobile forecast

irrationalagent:
Sorry I couldn't figure out what was going on with burnham import from looker-hub, it seemed like I was even with production. So I just changed things so I could put up the PR for now. Apparently there's something going on here I don't understand. 

I did notice that the main branch of spoke tries to pull from a burnham/dashboards directory from the hub project, but there is no such directory on the main branch of hub.

Also, I noticed that the change to use arrays was causing bigquery to return a "rows too large" error, I *think* because there are too many possible values for `app_name`, i.e. too many forecasts being generated. So I hard-coded a filter on mobile_usage_fields to only return the app_names we care about. that seemed to resolve the problem, but i guess its another  argument for forecasting one app_name at a time based on a filter rather than using `time_series_id_col` to do all the forecasts at once.


fbertsch (approved):
LGTM. The burnham issue is because of the automated deploys. I think it's fine to update the model file as you have done.

PR #41, opened 2021-04-29
Numbers that matter dashboard

scholtzan:
Migrated from dev: https://mozilladev.cloud.looker.com/dashboards-next/3?Date%20Range=28%20day&Countries=Overall&Shift%20N%20Days=8

It seems model hasn't been configured in Looker, and I don't have permissions to configure it.

This dashboard will replace https://protosaur.dev/numbers-that-matter/.

Some caveats:
* To provide a visualization of the funnel in the dashboard, the explores are combined using "Merge Result". Creating a single funnel explore by joining views is not possible. The views need to be aggregated first before they can be joined (which is essentially what "Merge Result" does) since , for example, the `ga_derived` cannot be joined with ping tables.
* The Numbers That Matter dashboard compares the numbers from the current time period with the time period before that. E.g. the numbers of the past 28 days are compared to the numbers of the 28 days before that. I added a `previous_time_period` parameter to indicate that data from the previous time period should be used to compute the numbers.
* The dashboard does by default not show the numbers of the most recent 8 days since computing the activation always lags 7 days behind


duet/explores/desktop_activation.explore.lkml
```diff
@@ -0,0 +1,46 @@
...
+    AND
+    DATE(${submission_timestamp_date}) > DATE_SUB(
+      IF({% parameter desktop_activation.previous_time_period %},
+      -- if the data for the previous time period is requested,
+      -- shift dates by the date range provided via the 'date' filter
+      DATE(DATE_ADD(
+        DATE({% date_start desktop_activation.date %}), INTERVAL DATE_DIFF(DATE({% date_start desktop_activation.date %}), DATE({% date_end desktop_activation.date %}), DAY) DAY)),
+        DATE({% date_start desktop_activation.date %})),
+      -- if the most recent week is to be ignored, shift date range by 8 days
+      INTERVAL IF({% parameter desktop_activation.ignore_most_recent_week %}, 8, 0) DAY);;
```
scholtzan, line 34: `ignore_most_recent_week` is by default set to true since we don't want to show the most recent 8 days in the dashboard and instead shift the date range by 8 days. When users select the "Past 28 days" date range they expect to see the numbers of the 28 days up until 8 days ago. In some cases this flag needs to be set to false, for example if users want to look at a specific date range in the past. This behaviour matches that of the existing Numbers That Matter dashboard

duet/explores/desktop_activation.explore.lkml
```diff
@@ -0,0 +1,46 @@
...
+      -- if the most recent week is to be ignored, shift date range by 8 days
+      INTERVAL IF({% parameter desktop_activation.ignore_most_recent_week %}, 8, 0) DAY)
+    AND
+    DATE(${submission_timestamp_date}) > DATE_SUB(
+      IF({% parameter desktop_activation.previous_time_period %},
+      -- if the data for the previous time period is requested,
+      -- shift dates by the date range provided via the 'date' filter
+      DATE(DATE_ADD(
+        DATE({% date_start desktop_activation.date %}), INTERVAL DATE_DIFF(DATE({% date_start desktop_activation.date %}), DATE({% date_end desktop_activation.date %}), DAY) DAY)),
+        DATE({% date_start desktop_activation.date %})),
```
scholtzan, line 32: The date range used for computing the numbers depends on whether `previous_time_period` is set to true or false. It is false by default, but if set to true it will use the data of the previous time period. For example, if the `date` filter is set to a range from 2021-04-11 to 2021-04-17 then the previous time period is 2021-04-04 to 2021-04-10.

duet/explores/desktop_activation.explore.lkml
```diff
@@ -0,0 +1,46 @@
...
+      -- if the most recent week is to be ignored, shift date range by 8 days
+      INTERVAL IF({% parameter desktop_activation.ignore_most_recent_week %}, 8, 0) DAY);;
+  join: country_buckets {
+    type: cross
+    relationship: many_to_one
+    sql_where: ${country_buckets.code} = ${desktop_activation.normalized_country_code} ;;
+  }
```
scholtzan, line 39: Numbers that matter assigns countries to certain buckets. Some countries are assigned to multiple buckets, for example 'GB' is in the `Overall` and `tier-1` bucket.

fbertsch (commented):
A few comments and questions. I noticed the dashboard is a bit slow with all of these aggregations happening. I think it might be good to [enable aggregate awareness for this dashboard.](https://docs.looker.com/data-modeling/learning-lookml/aggregate_awareness#exact_match)

I also saw some errors when trying to load the dashboard, are you seeing these as well?

Screen Shot 2021-04-30 at 11 53 37 AM

duet/views/desktop_new_profile.view.lkml
```diff
@@ -0,0 +1,61 @@
+include: "//looker-hub/duet/views/*.view.lkml"
+
+view: desktop_new_profile {
+  extends: [new_profile]
+
+  dimension: submission_timestamp_date {
```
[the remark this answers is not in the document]
scholtzan: It does. Using that instead

duet/views/desktop_install.view.lkml
```diff
@@ -0,0 +1,50 @@
...
+    type: date
+  }
+
+  dimension: attribution {
+    hidden: yes
+    type: string
+    sql: IF(${TABLE}.attribution IS NULL, "Unknown", SPLIT(SPLIT(${TABLE}.attribution, '26ua%3D')[SAFE_OFFSET(1)], '%')[SAFE_OFFSET(0)]) ;;
+  }
+
+  dimension: join_date {
```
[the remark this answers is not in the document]
scholtzan: Yes, I dropped it. I was using it previously for some exploration, but it's not needed in the dashboard

duet/explores/desktop_activation.explore.lkml
```diff
@@ -0,0 +1,46 @@
...
+      SAFE.PARSE_DATE('%Y%m%d', SUBSTR(${build_id}, 0, 8)),
+      MONTH
+    ) <= 1 AND
+    ${os} = "Windows" AND
+    ${attribution_source} IS NOT NULL AND
+    ${distribution_id} IS NULL AND
+    ${attribution_ua} != "firefox" AND
+    ${startup_profile_selection_reason} = "firstrun-created-default" AND
+    DATE(${submission_timestamp_date}) <= DATE_SUB(
+      IF({% parameter desktop_activation.previous_time_period %},
```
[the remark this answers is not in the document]
scholtzan: It's used in the dashboard but not exposed as a filter. It is used for the tiles on top of the page to calculate the change compared to the previous time period.

duet/explores/session.explore.lkml
```diff
@@ -0,0 +1,51 @@
+include: "../views/*.view.lkml"
+
+explore: session {
+  sql_always_where:
```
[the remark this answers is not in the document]
scholtzan: I agree! Do we already have capabilities to generate something like this?

scholtzan:
I fixed the errors in the dashboard now that I was able to actually view it myself. I also added `aggregate_tables` to the explores which should speed things up.

duet/explores/session.explore.lkml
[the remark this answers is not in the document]
fbertsch: Hmm, clearly we do this in `lookml-generator`, but given it relies on changes in `looker-spoke-default` I don't see a good path to exposing the explores in `looker-hub`.

We could consider, longer-term, supporting more generated code in `looker-spoke-default`, like we do for the generated models and directories. Then you might make some manual changes to views, add add the generated explores via lookml-generator. I'll file an issue there.

duet/explores/desktop_activation.explore.lkml
```diff
@@ -0,0 +1,74 @@
...
+      desktop_activation.date: "28 day",
+      desktop_activation.ignore_most_recent_week: "yes"
+    ]
+  }
+
+  aggregate_table: rollup__country_buckets_bucket__submission_date {
+    query: {
+      dimensions: [country_buckets.bucket, desktop_activation.submission_timestamp_date]
+      measures: [desktop_activation.activations]
+      filters: [desktop_activation.date: "28 days", desktop_activation.ignore_most_recent_week: "Yes"]
```
fbertsch, line 51: It doesn't look like Looker is using these because the dashboard filters on `"28 day"`, while this has `"28 days"`. Did you use the "generated aggregate table LookML" feature? If you did, there must be a bug in the implementation.

Try switching to `"28 day"` and we can check to see if the aggregate tables are used.

FYI, I checked by looking at the running queries in Looker. I saw this in the issued query:
```
-- Did not use desktop_activation::rollup__country_buckets_bucket__submission_date; filter field desktop_activation.date did not have the same filter value in the query ('28 day') as in the aggregate_table ('28 days')
-- Did not use desktop_activation::rollup__country_buckets_bucket__submission_date_prev; filter field desktop_activation.date did not have the same filter value in the query ('28 day') as in the aggregate_table ('28 days')
```

Now, you can do the same by inspecting queries from the BQ console. These queries all run in the mozdata project, so you can look through and find them to see if Looker is using the aggregate tables or the cache.

duet/explores/desktop_activation.explore.lkml
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    materialization: {
+      sql_trigger_value: SELECT CURRENT-DATE;;
```
fbertsch, line 55: Is this valid BQ SQL?

duet/explores/desktop_activation.explore.lkml
scholtzan: I fixed the filter and looked into the BigQuery logs. It's now using the cached tables

duet/explores/desktop_activation.explore.lkml
scholtzan: oops

fbertsch (approved):
r+, this is great, the dashboard is lightning fast now!

My only question is the explores. I'll let you decide how to handle them, and we can always iterate on them (but acknowledge that it may break people's dashboards, if they build something off of one). basically:
1. Do we want to expose all 5 of them to end users? Are they all useful as they are? Should we expose a single explore that somehow combines the various data? If not that, maybe a subset?
2. Do we want to name them more similarly? See below. Most of them start with "Desktop" but two don't.
3. We should probably add descriptions to them. I've also found that [example queries](https://docs.looker.com/reference/explore-params/query) can be useful for launching off points (when people come to explores directly). 
Screen Shot 2021-04-30 at 5 22 54 PM

scholtzan:
1. Actually, Releases doesn't need to be exposed. Having a single explore would be nice, but I wasn't able to combine them in a good way since the views need to be aggregated first before they can be joined since, for example, the `session` ( based on a `ga_derived` table) cannot be joined with ping tables. But also the separate explores are useful for investigations.
2. Releases will also be used for the Mobile Numbers That Matter Dashboard. I guess re-naming Session to Desktop Session does make sense.
3. Agreed.

fbertsch:
>     1. Actually, Releases doesn't need to be exposed. Having a single explore would be nice, but I wasn't able to combine them in a good way since the views need to be aggregated first before they can be joined since, for example, the `session` ( based on a `ga_derived` table) cannot be joined with ping tables. But also the separate explores are useful for investigations.
> 
>     2. Releases will also be used for the Mobile Numbers That Matter Dashboard. I guess re-naming Session to Desktop Session does make sense.
> 
>     3. Agreed.

Got it. I do wonder if it would have been better to expose a single explore that has all the NtM measures and the few dimensions it does offer (I remember us discussing exactly this now), but like we said we can iterate to that.

PR #42, opened 2021-05-04
Add `client_counts` explore for client-counting-shaped analyses

fbertsch:
This is largely to support cohort analyses, see the first example query in the explore. Using a table calc gets us to a more traditional cohort retention curve, see below (since we need the max of each column).

Jeff, I thought you might be interested in reviewing this, so we can begin to think about good ways to utilize the `_daily` and `_last_seen` explores more fully.

Screen Shot 2021-05-03 at 9 29 58 PM

### DIFF
--- a/firefox_ios/firefox_ios.model.lkml
+++ b/firefox_ios/firefox_ios.model.lkml
@@ -51,7 +51,7 @@ explore: client_counts {
   description: "Client counts across dimensions and cohorts."
   always_filter: {
     filters: [
-      client_counts.submission_date: "4 weeks ago",
+      client_counts.submission_date: "4 weeks",
       client_counts.channel: "mozdata.org^_mozilla^_ios^_firefox.baseline^_clients^_daily",
     ]
   }
@@ -64,8 +64,8 @@ explore: client_counts {
     measures: [client_counts.client_count]
     pivots: [client_counts.first_seen_week]
     filters: [
-      submission_date: "8 weeks ago",
-      first_seen_date: "8 weeks ago",
+      submission_date: "8 weeks",
+      first_seen_date: "8 weeks",
       client_counts.have_completed_period: "yes",
     ]
     sorts: [
--- a/firefox_ios/firefox_ios.model.lkml
+++ b/firefox_ios/firefox_ios.model.lkml
@@ -46,3 +46,45 @@ explore: counters {
     ]
   }
 }
+
+explore: client_counts {
+  description: "Client counts across dimensions and cohorts."
+  always_filter: {
+    filters: [
+      client_counts.submission_date: "4 weeks ago",
+      client_counts.channel: "mozdata.org^_mozilla^_ios^_firefox.baseline^_clients^_daily",
+    ]
+  }
+  query: cohort_analysis {
+    description: "Client Counts of weekly cohorts over the N days."
+    dimensions: [
+      client_counts.days_since_first_seen,
+      client_counts.first_seen_week
+    ]
+    measures: [client_counts.client_count]
+    pivots: [client_counts.first_seen_week]
+    filters: [
+      submission_date: "8 weeks ago",
+      first_seen_date: "8 weeks ago",
+      client_counts.have_completed_period: "yes",
+    ]
+    sorts: [
+      client_counts.days_since_first_seen: asc
+    ]
+  }
+  query: build_breakdown {
+    description: "Number of clients per build."
+    dimensions: [
+      client_counts.submission_date,
+      client_counts.app_build
+    ]
+    measures: [client_count]
+    pivots: [app_build]
+    filters: [
+      submission_date: "28 days"
+    ]
+    sorts: [
+      submission_date: asc
+    ]
+  }
+}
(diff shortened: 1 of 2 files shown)


firefox_ios/views/client_counts.view.lkml
```diff
@@ -0,0 +1,60 @@
...
+        {% if client_counts.days_since_first_seen._is_selected %}
+          INTERVAL ${client_counts.days_since_first_seen} DAY
+        {% elsif client_counts.weeks_since_first_seen._is_selected %}
+          INTERVAL ${client_counts.weeks_since_first_seen} WEEK
+        {% elsif client_counts.months_since_first_seen._is_selected %}
+          INTERVAL ${client_counts.months_since_first_seen} MONTH
+        {% elsif client_counts.years_since_first_seen._is_selected %}
+          INTERVAL ${client_counts.months_since_first_seen} YEAR
+        {% endif %}
+      ) < current_date
```
fbertsch, line 44: I know this is a real beast. But let's start with an example.

If we're cohorting on weeks, and we want to see, each week from cohort creation, what fraction of the cohort is still active. Then we would not want to include data for week 1 from last week's cohort, since this week hasn't finished - they would look artificially deflated in retention.

For that case, we have:
```
DATE_ADD(
  -- The following gives the first date after the end of this cohort's first_week
  DATE_ADD(DATE(${client_counts.first_seen_week}), INTERVAL 1 WEEK),
  -- Then we add the current period, e.g. are we looking two weeks in the future?
  INTERVAL ${client_counts.weeks_since_first_seen} WEEK
)
-- Now we check to see if we have the data for that period
< current_date
```

fbertsch: This would actually be a nice time to use that `last_etl_date` UDF, since there will be a few hours this is not correct, while ETL finishes before the day starts.

firefox_ios/firefox_ios.model.lkml
```diff
@@ -46,3 +46,45 @@ explore: counters {
...
+explore: client_counts {
+  description: "Client counts across dimensions and cohorts."
+  always_filter: {
+    filters: [
+      client_counts.submission_date: "4 weeks ago",
+      client_counts.channel: "mozdata.org^_mozilla^_ios^_firefox.baseline^_clients^_daily",
+    ]
+  }
+  query: cohort_analysis {
+    description: "Client Counts of weekly cohorts over the N days."
```
jklukas, line 59:
```suggestion
    description: "Client Counts of weekly cohorts over the past N days."
```

firefox_ios/views/client_counts.view.lkml
```diff
@@ -0,0 +1,60 @@
...
+      Indicates whether the cohort for this row have all had a chance to complete this interval.
+      For example, new clients from yesterday have not all had a chance to send a ping for today."
+    sql:
+      DATE_ADD(
+        {% if client_counts.first_seen_date._is_selected %}
+          DATE_ADD(DATE(${client_counts.first_seen_date}), INTERVAL 1 DAY)
+        {% elsif client_counts.first_seen_week._is_selected %}
+          DATE_ADD(DATE(${client_counts.first_seen_week}), INTERVAL 1 WEEK)
+        {% elsif client_counts.first_seen_month._is_selected %}
+          DATE_ADD(DATE(${client_counts.first_seen_month}), INTERVAL 1 MONTH)
```
jklukas, line 30: I tried to pivot on first_seen_month and got an error about `2021-03` being an invalid timestamp format.

firefox_ios/views/client_counts.view.lkml
```diff
@@ -0,0 +1,60 @@
...
+      Only for use with cohort analysis.
+      Filter on true to remove the tail of incomplete data from cohorts.
+      Indicates whether the cohort for this row have all had a chance to complete this interval.
+      For example, new clients from yesterday have not all had a chance to send a ping for today."
+    sql:
+      DATE_ADD(
+        {% if client_counts.first_seen_date._is_selected %}
+          DATE_ADD(DATE(${client_counts.first_seen_date}), INTERVAL 1 DAY)
+        {% elsif client_counts.first_seen_week._is_selected %}
+          DATE_ADD(DATE(${client_counts.first_seen_week}), INTERVAL 1 WEEK)
```
jklukas, line 28: When I pivot by first_seen_week, I see only one line plotted, corresponding to the earliest available week. I don't understand why I don't see other lines.

jklukas: Your example Look works: https://mozilla.cloud.looker.com/looks/9

It looks like the filters there are "in the past" whereas the example cohort analysis here is using "matches (advanced)".

fbertsch: Okay this one was an error on my end - I changed the filters to use "ago", which I thought would be the same as "x completed weeks", but instead it's an exact week that is 8 weeks ago. Changing it to "in the past 8 weeks" shows the correct data. I'll update the filters.

firefox_ios/views/client_counts.view.lkml
fbertsch: Hmm, I didn't realize Looker would truncate the date string like that. Let's see how we can handle it.

fbertsch: Okay, these are fixed now.

jklukas (approved):
This all appears usable now with the fixes. Looks good to me.

PR #50, opened 2021-05-12

dataops-looker:


fbertsch (changes requested):
Hold off on merging this - I need to test some role set config on prod, see https://github.com/mozilla/lookml-generator/issues/95.

fbertsch (approved):
Should be fine to merge and let https://github.com/mozilla/lookml-generator/pull/105 land after.

PR #53, opened 2021-05-19
add total mobile kpi metrics

irrationalagent:
ok so this turned out to be way easier than i realized.... i need a vacation. ill close the other PR

### DIFF
--- a/KPI/dashboards/corporate_kpis.dashboard.lookml
+++ b/KPI/dashboards/corporate_kpis.dashboard.lookml
@@ -1,0 +1,744 @@
+- dashboard: firefox_corporate_kpis
+  title: Firefox Corporate KPIs
+  layout: newspaper
+  preferred_viewer: dashboards-next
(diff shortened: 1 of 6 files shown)


irrationalagent:
@ethompso28 switch to the `new_topline_dash` branch and you can view the changes to the `mobile_kpi_dashboard.dashboard` file that way

irrationalagent:
@ethompso28 I added a top level KPI dashboard per your spec in slack. its defined in the `corporate_kpis.dashboard` file.

one thing that's not quite as you envisioned is that right now the quarterly tables only show results to date, they do not show the forecasted expectations for future quarters since the forecasts are (the right side of a) left joined onto the actual data. we'll need to explore a way to get those to show up without changing too much.

again you have to switch to the `new_topline_dash` branch and pull from remote to view it properly.

ethompso28 (approved):
r+

PR #54, opened 2021-05-21
[DS-1545] View, explore, and dashboard for ios numbers that matter

acmiyaguchi:
This is a followup to #52 for [DS-1545](https://jira.mozilla.com/browse/DS-1545). This will follow the same general shape as the dashboard there.

This was actually straightforward to implement since the dashboard definition was easy to edit from android to ios. The metric definitions are different in the apple store than they are in the play store, so this will need to be documented.

### DIFF
--- a/duet/explores/mobile_ios_country.explore.lkml
+++ b/duet/explores/mobile_ios_country.explore.lkml
@@ -2,7 +2,8 @@ include: "../views/*.view.lkml"
 
 explore: mobile_ios_country {
   label: "Mobile iOS Funnel"
-  description: "Consolidated view of the Android acquisition funnel by country."
+  view_label: "iOS Funnel Steps"
+  description: "Consolidated view of the iOS acquisition funnel by country."
   join: country_buckets {
     type: inner
     relationship: many_to_one
--- a/duet/explores/mobile_ios_country.explore.lkml
+++ b/duet/explores/mobile_ios_country.explore.lkml
@@ -11,4 +11,73 @@ explore: mobile_ios_country {
       country_buckets.bucket: "Overall"
     ]
   }
+
+  aggregate_table: rollup__submission_date__0 {
+    query: {
+      dimensions: [
+        # "app_id" is filtered on in the dashboard.
+        # Uncomment to allow all possible filters to work with aggregate awareness.
+        # app_id,
+        # "country_buckets.bucket" is filtered on in the dashboard.
+        # Uncomment to allow all possible filters to work with aggregate awareness.
+        # country_buckets.bucket,
+        # "history_days" is filtered on in the dashboard.
+        # Uncomment to allow all possible filters to work with aggregate awareness.
+        # history_days,
+        submission_date
+      ]
+      measures: [activated, first_seen, first_time_installs, installations_opt_in, product_page_views]
+      filters: [
+        # "country_buckets.bucket" is filtered on by the dashboard. The filter
+        # value below may not optimize with other filter values.
+        country_buckets.bucket: "tier-1",
+        # "mobile_ios_country.app_id" is filtered on by the dashboard. The filter
+        # value below may not optimize with other filter values.
+        mobile_ios_country.app_id: "firefox",
+        # "mobile_ios_country.history_days" is filtered on by the dashboard. The filter
+        # value below may not optimize with other filter values.
+        mobile_ios_country.history_days: "7"
+      ]
+    }
+
+    # Please specify a datagroup_trigger or sql_trigger_value
+    # See https://looker.com/docs/r/lookml/types/aggregate_table/materialization
+    materialization: {
+      sql_trigger_value: SELECT CURRENT_DATE();;
+    }
+  }
+
+  aggregate_table: rollup__activated__first_seen__first_time_installs__installations_opt_in__product_page_views__1 {
+    query: {
+      dimensions: [
+        # "app_id" is filtered on in the dashboard.
+        # Uncomment to allow all possible filters to work with aggregate awareness.
+        # app_id,
+        # "country_buckets.bucket" is filtered on in the dashboard.
+        # Uncomment to allow all possible filters to work with aggregate awareness.
+        # country_buckets.bucket,
+        # "history_days" is filtered on in the dashboard.
+        # Uncomment to allow all possible filters to work with aggregate awareness.
+        # history_days
+      ]
+      measures: [activated, first_seen, first_time_installs, installations_opt_in, product_page_views]
+      filters: [
+        # "country_buckets.bucket" is filtered on by the dashboard. The filter
+        # value below may not optimize with other filter values.
+        country_buckets.bucket: "tier-1",
+        # "mobile_ios_country.app_id" is filtered on by the dashboard. The filter
+        # value below may not optimize with other filter values.
+        mobile_ios_country.app_id: "firefox",
+        # "mobile_ios_country.history_days" is filtered on by the dashboard. The filter
+        # value below may not optimize with other filter values.
+        mobile_ios_country.history_days: "7"
+      ]
+    }
+
+    # Please specify a datagroup_trigger or sql_trigger_value
+    # See https://looker.com/docs/r/lookml/types/aggregate_table/materialization
+    materialization: {
+      sql_trigger_value: SELECT CURRENT_DATE();;
+    }
+  }
 }
--- a/duet/explores/mobile_ios_country.explore.lkml
+++ b/duet/explores/mobile_ios_country.explore.lkml
@@ -1,5 +1,14 @@
 include: "../views/*.view.lkml"
 
 explore: mobile_ios_country {
-
+  join: country_buckets {
+    type: inner
+    relationship: many_to_one
+    sql_on:  ${country_buckets.code} = ${mobile_ios_country.country} ;;
+  }
+  always_filter: {
+    filters: [
+      country_buckets.bucket: "Overall"
+    ]
+  }
 }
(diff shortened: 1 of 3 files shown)


duet/views/mobile_ios_country.view.lkml
```diff
@@ -0,0 +1,368 @@
...
+  }
+
+  measure: first_seen {
+    description: "The number of client ids seen for the first time in the baseline clients last seen table."
+    type: sum
+    sql: ${TABLE}.first_seen ;;
+  }
+
+  measure: activated {
+    description: "The number of clients that have used the app for 5/7 days."
```
fbertsch, line 350:
```suggestion
    description: "The number of clients that have used the app for at least 5 of their first 7 days."
```

duet/views/mobile_ios_country.view.lkml
```diff
@@ -0,0 +1,368 @@
...
+  }
+
+  measure: installations_opt_in {
+    description: "The number of opt-in first time runs reported by the apple store"
+    type: sum
+    sql: ${TABLE}.installations_opt_in ;;
+  }
+
+  measure: first_seen {
+    description: "The number of client ids seen for the first time in the baseline clients last seen table."
```
fbertsch, line 344:
```suggestion
    description: "The number of client ids that sent a baseline ping (and accounted for at least 1 DOU)."
```

duet/views/mobile_ios_country.view.lkml
```diff
@@ -0,0 +1,368 @@
...
+    allowed_value: {
+      value: "0"
+    }
+    allowed_value: {
+      value: "1"
+    }
+  }
+
+  dimension: bucket {
+    description: "The geographical bucket that the lies under. Do not combine with country unless this is set to overall."
```
fbertsch, line 276:
```suggestion
    description: "The geographical bucket that the country lies under. Do not combine with country unless this is set to overall."
```

duet/views/mobile_ios_country.view.lkml
```diff
@@ -0,0 +1,368 @@
...
+    }
+    allowed_value: {
+      label: "Focus"
+      value: "focus"
+    }
+    allowed_value: {
+      label: "Klar (Focus)"
+      value: "klar"
+    }
+  }
```
fbertsch, line 242: Can we do something like `app_name` here instead? `app_id` may be confusing.
```suggestion
  parameter: app_name {
    description: "The name of the application."
    type:  unquoted
    default_value: "firefox"
    allowed_value: {
      label: "Firefox for iOS"
      value: "firefox"
    }
    allowed_value: {
      label: "Focus for iOS"
      value: "focus"
    }
    allowed_value: {
      label: "Klar (Focus)"
      value: "klar"
    }
  }
```

duet/explores/mobile_ios_country.explore.lkml
```diff
@@ -0,0 +1,85 @@
+include: "../views/*.view.lkml"
+
+explore: mobile_ios_country {
+  label: "Mobile iOS Funnel"
+  description: "Consolidated view of the Android acquisition funnel by country."
```
fbertsch, line 5:
```suggestion
  description: "Consolidated view of the iOS acquisition funnel by country."
```

duet/explores/mobile_ios_country.explore.lkml
```diff
@@ -0,0 +1,85 @@
+include: "../views/*.view.lkml"
+
+explore: mobile_ios_country {
+  label: "Mobile iOS Funnel"
```
fbertsch, line 4:
```suggestion
  label: "Mobile iOS Funnel"
  view_label: "iOS Funnel Steps"
```

that would apply to this name here:
Screen Shot 2021-05-24 at 10 09 43 AM

duet/views/mobile_ios_country.view.lkml
```diff
@@ -0,0 +1,368 @@
...
+
+  dimension: latest_date {
+    description: "The most recent submission date used for analysis"
+    type: date
+    datatype: date
+    sql: ${TABLE}.latest_date ;;
+  }
+
+  measure: product_page_views {
```
fbertsch, line 325: What are your thoughts on putting `step_N` before each of these? That would make it clearer what the funnel actually is in the explore, and would then also order the measures by their step count (rather than alpha by name).

fbertsch (approved):
r+wc

Looking even less funnelly than Android :(  How is Apple reporting more first-time runs than first-time installs? I would say it's historical installs that are finally running Fx, but the count is >2x, which would be unsustainable.

duet/views/mobile_ios_country.view.lkml
acmiyaguchi: This is slightly inconsistent with the android view. Should I modify this there, too?

duet/views/mobile_ios_country.view.lkml
acmiyaguchi: It might not be a bad idea, though the name of the measure will get somewhat long. This would probably go in a follow up so the android view is consistent in naming.

acmiyaguchi:
> How is Apple reporting more first-time runs than first-time installs? 

I'm not sure on the details of how the opt-in installation metric is reported, but there is the mechanism where iOS will uninstall your apps for you to save on space. This *could* be treated as a first-run (installation opt-in), but not a first time install.

duet/views/mobile_ios_country.view.lkml
fbertsch: I think that makes sense, just for greater clarity. I wouldn't expect everyone who's using the explores to know what an `app_id` is.

fbertsch:
> > How is Apple reporting more first-time runs than first-time installs?
> 
> I'm not sure on the details of how the opt-in installation metric is reported, but there is the mechanism where iOS will uninstall your apps for you to save on space. This _could_ be treated as a first-run (installation opt-in), but not a first time install.

Okay - I may recommend adding a commentary to the dashboard itself, to outline clear issues like this (as well as the high `first_run` count).

acmiyaguchi:
Yes, documentation on the dashboard is certainly planned, with @ksiegler1 to iron out some of the details about how the measures are reported.

duet/views/mobile_ios_country.view.lkml
```diff
@@ -0,0 +1,368 @@
...
+            THEN
+              'SZ'
+            WHEN
+              name = 'Laos'
+            THEN
+              'LA'
+            ELSE
+              code
+            END
+            AS country,
```
fbertsch, line 145: That static dataset doesn't have these?

duet/views/mobile_ios_country.view.lkml
```diff
@@ -0,0 +1,368 @@
...
+            date,
+            app_name,
+            CASE
+            WHEN
+              storefront LIKE '%St.%'
+            THEN
+              REPLACE(storefront, 'St.', 'Saint')
+            WHEN
+              storefront = 'China mainland'
+            THEN
```
fbertsch, line 40: What do you think about pulling this big thing out in to a UDF? That would simplify things here, and make it usable for other analyses that need to join apple and telemetry data.

duet/views/mobile_ios_country.view.lkml
acmiyaguchi: The apple store doesn't have country code, only country names (for the store front). They also use different names in a few dozen cases that have to be manually mapped e.g. China Mainland. This is mapping the country name to the static dataset.

I believe those countries are missing from the static dataset, although @ksiegler1 worked through this logic for the most part.

duet/views/mobile_ios_country.view.lkml
acmiyaguchi: That seems fine, the logic is kind of wonky. I didn't want to go through the hassle when I was building this since it would have taken longer to get to the end state (getting the udf deployed from bigquery-etl).

fbertsch: I understand, but it will give a better guarantee of correctness here. The existing query is hard to review because of the length.

acmiyaguchi: Okay, I filed https://github.com/mozilla/bigquery-etl/issues/2068.

duet/views/mobile_ios_country.view.lkml
ksiegler1: Yes these ones had to be manually added because I could not locate them in the static country codes dataset.

ksiegler1:
>  How is Apple reporting more first-time runs than first-time installs?

This could be for a couple of reasons. The first time installs metric (derived from `app_units`) has some caveats

- It's a measure of first time user downloads from the App store
- Re-downloads to other iOS devices from the same user are not counted
- The download action occurs in the Apple App store and falls outside the opt-in requirement, since it's not technically "device usage". So this data is reported regardless of user decision to opt in/out of data sharing.

On the other hand, for the `installations_opt_in` metrics

- It's a measure of first time **opening** of the app on a given device
- Requires the user to opt in to sharing data with Apple - probably the most important reason for the discrepancy

So if users download the app and never open it, we could see App Units > Installations. If they do open it but have opted out of sharing their data, we could see the same effect.

Like @acmiyaguchi mentioned this will all be part of documentation on the dashboard, since it doesn't appear to behave like a funnel.

duet/views/mobile_ios_country.view.lkml
acmiyaguchi: Resolved

acmiyaguchi:
I'm merging this in, with the expectation that documentation will follow before it is imported into the DUET folder.

PR #55, opened 2021-05-22
Add labels to android ntm explore and parameters

acmiyaguchi:
Follow up to #52

### DIFF
--- a/duet/explores/mobile_android_country.explore.lkml
+++ b/duet/explores/mobile_android_country.explore.lkml
@@ -1,7 +1,7 @@
 include: "../views/*.view.lkml"
 
 explore: mobile_android_country {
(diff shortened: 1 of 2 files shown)


duet/explores/mobile_android_country.explore.lkml
```diff
@@ -1,6 +1,8 @@
 include: "../views/*.view.lkml"
 
 explore: mobile_android_country {
+  label: "Mobile Android Funnel"
```
acmiyaguchi, line 4: I prefixed this with Mobile so that it will end up next to the iOS one. I also shortened the name so it fits on a single line.

fbertsch (approved):
LGTM

PR #56, opened 2021-05-24
Changes to Corporate KPI Dashboard

irrationalagent:
@ethompso28 I think I covered all the requested changes, please take a look and LMK if I missed anything
(switch to the `kpi-changes` branch and then view the dashboard from the `corporate_kpis.dashboard` lookml file).

### DIFF
--- a/KPI/dashboards/corporate_kpis.dashboard.lookml
+++ b/KPI/dashboards/corporate_kpis.dashboard.lookml
@@ -80,10 +80,10 @@
       determining if the actuals are truly starting to differ from the forecast. '
     listen:
       Date: firefox_desktop_usage_2021.date
-    row: 16
-    col: 0
-    width: 24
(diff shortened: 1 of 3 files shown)


ethompso28:
Only small nits:
1) Spell out "moving average" instead of MA
2) Add 2020 DAU MA for mobile?
3) make the text box bigger under mobile diff chart so that it doesnt do that mini scroll thing 
4) make titles the same for desktop and mobile burn up charts since they are now side by side
5) Make the main desktop and mobile charts one or two notches shorter for ease of use for screenshotting

irrationalagent:
OK, I think my latest commit addressed these

irrationalagent:
@ethompso28 if these look good to you, can you rubber stamp?

ethompso28 (approved):
r+

PR #58, opened 2021-05-25
[DS-1610] Make OS Version and App Version Numbers in iOS counterse

acmiyaguchi:
[jira](https://jira.mozilla.com/browse/DS-1610)

This adds two derived dimensions for sorting in the correct order. 

On another note, I tried adding a suggest explore but it didn't seem to do anything. Any ideas on why that might be?

### DIFF
--- a/firefox_ios/views/counters.view.lkml
+++ b/firefox_ios/views/counters.view.lkml
@@ -1,22 +1,3 @@
-explore: app_display_major_version {
-  hidden: yes
-}
-
-view: app_display_major_version {
-  derived_table: {
-    sql:
-      SELECT DISTINCT cast(split(app_display_version, ".")[offset(0)] as int64) as app_display_major_version
-      FROM `moz-fx-data-shared-prod`.org_mozilla_ios_firefox.metrics m
-      WHERE DATE(submission_timestamp) > DATE_SUB(current_date(), interval 14 day)
-      ORDER BY 1 ;;
-  }
-
-  dimension: app_display_major_version {
-    type: number
-    sql:  ${TABLE}.app_display_major_version ;;
-  }
-}
-
 view: counters {
   derived_table: {
     sql:
@@ -158,8 +139,6 @@ view: counters {
   dimension: app_display_major_version {
     type: number
     sql: cast(split(${TABLE}.app_display_version, ".")[offset(0)] as int64) ;;
-    suggest_explore: app_display_major_version
-    suggest_dimension: app_display_major_version.app_display_major_version
   }
 
   dimension: architecture {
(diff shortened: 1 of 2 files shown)


firefox_ios/views/counters.view.lkml
```diff
@@ -1,22 +1,3 @@
...
-      WHERE DATE(submission_timestamp) > DATE_SUB(current_date(), interval 14 day)
-      ORDER BY 1 ;;
-  }
-
-  dimension: app_display_major_version {
-    type: number
-    sql:  ${TABLE}.app_display_major_version ;;
-  }
-}
-
```
acmiyaguchi: @fbertsch When I tried the counter explore with this explore as a suggestion for the new derived field, it didn't seem to come up with any results. Any ideas on what I did wrong here? I'd like to use a similiar approach for the auto-generated labeled counters in the future...

fbertsch (approved):
This is definitely an improvement, but we should also enable minor versions (probably as a new dimension).

re: suggestions, Looker only suggests for string fields. I could see it being useful to filter on numeric fields but it isn't available.
Screen Shot 2021-05-24 at 8 53 37 PM

fbertsch (changes requested):
Ah shoot, we should add these to `unified metrics`, not `counters`, since we're going to [deprecate the latter](https://bugzilla.mozilla.org/show_bug.cgi?id=1711110).

acmiyaguchi:
I'll add this to the unified metrics explorer too 👍 

Counters is the explore being used in the issue; unified metrics only contains the subset of metrics that's common to legacy and glean. Once we have the automated explores generated, the underlying view will be useful for general purpose analysis.

firefox_ios/views/counters.view.lkml
```diff
@@ -137,6 +136,16 @@ view: counters {
     suggest_dimension: metrics_sample.app_display_version
   }
 
+  dimension: app_display_major_version {
+    type: number
+    sql: mozfun.norm.truncate_version(${app_display_version}, "major") ;;
+  }
+
```
fbertsch, line 143: Looks like `mozfun.norm.truncate(x, "minor")` will pull out the major and minor versions, so we probably only need one field with the minor version below.
```suggestion
```

fbertsch (approved):
r+wc

If we're going to add the minor version number fields, let's also hide the stringified fields. They will remain for historic queries but new ones will use the numeric fields.

firefox_ios/views/counters.view.lkml
acmiyaguchi: I don't think this is necessarily true, I think having just the major version is useful too.

fbertsch: I think I'd rather people get comfortable using more generic fields, but if you think it's good for this instance then we should keep it.

acmiyaguchi: I don't think there's a way to round a dimension, if you only wanted to see major version over time. The major-minor versions can be too noisy.

PR #63, opened 2021-05-27
Add measures for Activity Stream, Search, and Contextual Services

fbertsch:
This covers the tables we outlined in bug 1711955. We will probably follow this up with additional tables for both CS and AS.

### DIFF
--- a/search/views/search_clients_engine_sources_daily.view.lkml
+++ b/search/views/search_clients_engine_sources_daily.view.lkml
@@ -1,0 +1,58 @@
+include: "//looker-hub/search/views/search_clients_engines_sources_daily.view.lkml"
+
+view: +search_clients_engines_sources_daily {
+  measure: total_searches {
+    type: sum
+    sql: ${sap} ;;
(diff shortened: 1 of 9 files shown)


search/views/search_clients_engine_sources_daily.view.lkml
```diff
@@ -0,0 +1,58 @@
+include: "//looker-hub/search/views/search_clients_engines_sources_daily.view.lkml"
+
+view: +search_clients_engines_sources_daily {
```
fbertsch, line 3: These measures I added deserve a close eye from the search team:
- Do these names make sense for what the query is doing?
- Are the descriptions correct?
- Do we need to remove some or add more variants?

Note closely what the UI gives you in the info field, which does include the SQL fields being summed (those will not, however, be available to see in visualizations):
Screen Shot 2021-05-26 at 11 21 09 PM

mmccorks: This looks good to me.

mmccorks (approved):
LGTM

fbertsch:
Thank you @mmccorks!

PR #66, opened 2021-05-31
Numbers that matter dashboard review

scholtzan:
Addresses feedback from https://docs.google.com/document/d/1spwKE5C1LrUo4D5kak6JPOoRy55CDQUA4giiqiA3QTQ/edit#


duet/explores/desktop_activation.explore.lkml
```diff
@@ -21,8 +21,8 @@ explore: desktop_activation {
         DATE(DATE_ADD(
           DATE({% date_end desktop_activation.date %}), INTERVAL DATE_DIFF(DATE({% date_start desktop_activation.date %}), DATE({% date_end desktop_activation.date %}), DAY) DAY)),
           DATE({% date_end desktop_activation.date %})),
-      -- if the most recent week is to be ignored, shift date range by 8 days
-      INTERVAL IF({% parameter desktop_activation.ignore_most_recent_week %}, 8, 0) DAY)
+      -- if the most recent week is to be ignored, shift date range by 9 days
```
acmiyaguchi, line 24: What's the motivation behind changing from 8 days to 9 days?

duet/views/desktop_activation.view.lkml
```diff
@@ -10,31 +10,65 @@ view: desktop_activation {
...
+            DATE(submission_timestamp) > DATE_SUB(
+              IF({% parameter desktop_activation.previous_time_period %},
+                -- if the data for the previous time period is requested,
+                -- shift dates by the date range provided via the 'date' filter
+                DATE(DATE_ADD(
+                  DATE({% date_start desktop_activation.date %}),
+                  INTERVAL DATE_DIFF(DATE({% date_start desktop_activation.date %}), DATE({% date_end desktop_activation.date %}), DAY) DAY)),
+                DATE({% date_start desktop_activation.date %})),
+              -- if the most recent week is to be ignored, shift date range by 9 days
+              INTERVAL IF({% parameter desktop_activation.ignore_most_recent_week %}, 9, 0) DAY)
```
acmiyaguchi, line 63: Consider making this a sql udf (though I'm not sure if it can be used as a partition filter). Something like:

```
valid_date_range(
  {% date_start desktop_activation.date %}, 
  {% date_end desktop_activation.date %},
  {% parameter desktop_activation.ignore_most_recent_week %},
  9
)

File: duet/explores/desktop_new_profile.explore.lkml [hunk 3]
```diff
@@ -1,5 +1,57 @@
...
+        -- if the data for the previous time period is requested,
+        -- shift dates by the date range provided via the 'date' filter
+        DATE(DATE_ADD(
+          DATE({% date_start desktop_new_profile.date %}),
+          INTERVAL DATE_DIFF(DATE({% date_start desktop_new_profile.date %}), DATE({% date_end desktop_new_profile.date %}), DAY) DAY)),
+        DATE({% date_start desktop_new_profile.date %})),
+      -- if the most recent week is to be ignored, shift date range by 9 days
+      INTERVAL IF({% parameter desktop_new_profile.ignore_most_recent_week %}, 9, 0) DAY) AND
+    payload.processes.parent.scalars.startup_profile_selection_reason = 'firstrun-created-default';;
+  }
```
acmiyaguchi, line 35: Why not get the most recent row number in the view instead of exposing the row number and filtering into the explore? Are there other uses for the row number?

acmiyaguchi (approved):
This addresses my feedback 👍 

image

Some questions inline, but otherwise it looks reasonable.

duet/explores/desktop_new_profile.explore.lkml
scholtzan: The row number is used to ensure we count duplicate entries only once. Looker doesn't support aggregate function in dimension definitions. So having `ROW_NUMBER() OVER (PARTITION BY client_id, DATE(submission_timestamp))` as a dimension in `desktop_new_profile` is not possible. 

It would have been possible to rewrite `desktop_new_profile` as a `derived_table`, however then taking advantage of the automatically generated view via lookml-generator isn't possible anymore.

Having a separate view with distinct clients used to ensure we don't have duplicates seemed the best way forward here.

duet/explores/desktop_new_profile.explore.lkml
```diff
@@ -1,5 +1,57 @@
...
+
+  dimension: client_id {
+    hidden: yes
+    type: string
+    sql: ${TABLE}.client_id ;;
+  }
+
+  dimension: rn {
```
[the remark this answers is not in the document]
scholtzan: Yes. But actually it's not really needed anywhere, so I updated it and set `hidden: yes`

duet/explores/desktop_activation.explore.lkml
scholtzan: The 7 day activation can lag 9 days behind, then the value in the graph is set to 0 which might cause some confusion. Shifting by 9 days also seems to be what is done in the previous Numbers That Matter dashboard

SuYoungHong:
LGTM

PR #71, opened 2021-06-07
Add descriptions to contextual services explores

fbertsch:


contextual_services/contextual_services.model.lkml
```diff
@@ -2,3 +2,33 @@ connection: "bigquery-oauth"
...
+    The impression ping would only count the last QuickSuggest link if any when the search session ends.
+    This data is kept for 30 days."
+}
+
+explore: +topsites_clicks {
+  description: "A topsites click ping is sent when a user clicks on a sponsored top site. This data is kept for 30 days."
+}
+
+explore: +topsites_impressions {
+  description: "A topsites impression ping is sent on impressions of sponsored topsites. This data is kept for 30 days."
```
fbertsch, line 33: There is not much in the contextual services doc about the topsites pings. I'm not sure if this is even correct, i.e. are we only sending this for sponsored topsites impressions?

contextual_services/contextual_services.model.lkml
```diff
@@ -2,3 +2,33 @@ connection: "bigquery-oauth"
...
+}
+
+explore: +quicksuggest_impressions {
+  description: "
+    This impression ping is only sent on two conditions:
+    A user completes the search session in the URL bar, i.e. he or she navigates away from the search bar, or
+    a QuickSuggest link was shown in the URL bar at the end of the search session.
+    This means that any QuickSuggest link(s) that was displayed while the user was still typing would not trigger this impression ping.
+    The impression ping would only count the last QuickSuggest link if any when the search session ends.
+    This data is kept for 30 days."
```
rebecca-burwei, line 25: Could we change this to:

 A user completes the search session in the URL bar, i.e. he or she navigates away from the search bar, AND
 a QuickSuggest link was shown in the URL bar at the end of the search session.
 This means that any QuickSuggest link(s) that were displayed while the user was still typing would not trigger this impression ping.
This data is kept for 30 days."

Changes:

- Line 21: AND condition instead of OR
- Line 23: Small grammar fix
- Remove line 24: It's saying the same thing as line 23, but not resilient to the possibility that, in the future, we might show more than one QuickSuggest link at the same time.

contextual_services/contextual_services.model.lkml
```diff
@@ -2,3 +2,33 @@ connection: "bigquery-oauth"
...
+
+explore: +event_aggregates {
+  description: "Daily event rollups. This aggregated dataset is faster to query and contains a longer history."
+}
+
+explore: +quicksuggest_clicks {
+  description: "
+    The QuickSuggest click ping is sent when the user clicks on the QuickSuggest link.
+    It shares most of the fields as the QuickSuggest impression ping except that it doesn’t include the user search query, nor the matched keywords.
+    This data is kept for 30 days."
```
rebecca-burwei, line 15: Can we add this sentence: Click pings may be sent more frequently than impression pings.

contextual_services/contextual_services.model.lkml
rebecca-burwei: These two topsites tables only contain impressions and clicks for sponsored topsites. Pings are sent for other types of topsites, but that data is not included in these two tables. A sanity check for this is that you don't see any rows in those tables with missing data for the advertiser field.

rebecca-burwei (changes requested):

contextual_services/contextual_services.model.lkml
fbertsch: Can you describe how? Is this because of multiple clicks for the same impression, or because we're somehow not sending an impression ping?

contextual_services/contextual_services.model.lkml
rebecca-burwei: In my understanding of the two conditions, the second one is a subset of the first one, but I will double check with Nan.

contextual_services/contextual_services.model.lkml
rebecca-burwei: We send separate pings for clicks and impressions to facilitate data processing and data sharing with our partners. More details here: https://docs.google.com/document/d/1crGV6l9wXh2ITmvj6tC5taBGM23XcUaf4B89w_cPdwE/edit#bookmark=id.g5v9vhy7sfry

contextual_services/contextual_services.model.lkml
rebecca-burwei: Confirmed from Nan that it's AND

contextual_services/contextual_services.model.lkml
fbertsch: Okay, I was misunderstanding. You mean they will be sent with lower latency, is that right?

fbertsch:
@rebecca-burwei I updated the descriptions, can you give this another look?

contextual_services/contextual_services.model.lkml
rebecca-burwei: yes, sorry, latency not frequency.

rebecca-burwei (approved):

PR #80, opened 2021-06-11
Add script to sync user-defined dashboards with LookML dashboards

jasonthomas:
For https://bugzilla.mozilla.org/show_bug.cgi?id=1713204.

I've tested this a bit on dev and works as expected. I am assuming we only want to link and sync dashboards on our 'production' instance, but we could potentially do it for all our instances.

I've tried to keep this minimal since we want to move this (and potentially other management type code) into another repo.

r?

### DIFF
--- a/lookml_dashboards.yaml
+++ b/lookml_dashboards.yaml
@@ -1,3 +1,14 @@
+# Listing of synced dashboards.
+# A synced dashboard is a User-Defined dashboard that
+# a  LookML Dashboard is copied to. It will match the
+# LookML Dashboard exactly.
+# 
+# Usage:
+#   1. Create a LookML dashboard.
+#   2. Create a User-Defined dashboard
+#   3. Below, add an entry syncing the two. Syncs will happen on merge to `main`.
+#
+# Format:
 # user_defined_dashboard_id: "lookml_dashboard_id"
 ---
 52: "kpi::firefox_corporate_kpis"
--- a/lookml_dashboards.yaml
+++ b/lookml_dashboards.yaml
@@ -12,4 +12,4 @@
 # user_defined_dashboard_id: "lookml_dashboard_id"
 ---
 101: "kpi::firefox_corporate_kpis"
-101: "duet::desktop_numbers_that_matter"
+66: "duet::desktop_numbers_that_matter"
--- a/lookml_dashboards.yaml
+++ b/lookml_dashboards.yaml
@@ -1,0 +1,4 @@
+# user_defined_dashboard_id: "lookml_dashboard_id"
+---
+52: "kpi::firefox_corporate_kpis"
+101: "duet::desktop_numbers_that_matter"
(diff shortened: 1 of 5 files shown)


lookml_dashboards.yaml
```diff
@@ -0,0 +1,4 @@
+# user_defined_dashboard_id: "lookml_dashboard_id"
```
fbertsch, line 1:
```suggestion
# Listing of synced dashboards.
# A synced dashboard is a User-Defined dashboard that
# a  LookML Dashboard is copied to. It will match the
# LookML Dashboard exactly.
# 
# Usage:
#   1. Create a LookML dashboard.
#   2. Create a User-Defined dashboard
#   3. Below, add an entry syncing the two. Syncs will happen on merge to `main`.
#
# Format:
# user_defined_dashboard_id: "lookml_dashboard_id"
```

src/sync_dashboards/main.py
```diff
@@ -0,0 +1,75 @@
...
+                    models.WriteDashboard(lookml_link_id=lookml_dashboard_id),
+                )
+            elif dashboard.lookml_link_id != lookml_dashboard_id:
+                logging.warning(
+                    f"Dashboard {dashboard_id} lookml_link_id changing from {dashboard.lookml_link_id} to {lookml_dashboard_id}"
+                )
+                sdk.update_dashboard(
+                    str(dashboard_id),
+                    models.WriteDashboard(lookml_link_id=lookml_dashboard_id),
+                )
```
fbertsch, line 46: Are there any downsides to updating the dashboards every time?
```suggestion
            sdk.update_dashboard(
                str(dashboard_id),
                models.WriteDashboard(lookml_link_id=lookml_dashboard_id),
            )
```

fbertsch (approved):
r+wc. I don't think we need to worry about dev/stage right now, since these changes should probably only happen in prod

lookml_dashboards.yaml
```diff
@@ -0,0 +1,15 @@
...
+# 
+# Usage:
+#   1. Create a LookML dashboard.
+#   2. Create a User-Defined dashboard
+#   3. Below, add an entry syncing the two. Syncs will happen on merge to `main`.
+#
+# Format:
+# user_defined_dashboard_id: "lookml_dashboard_id"
+---
+52: "kpi::firefox_corporate_kpis"
```
fbertsch, line 14:
```suggestion
101: "kpi::firefox_corporate_kpis"
```

lookml_dashboards.yaml
```diff
@@ -0,0 +1,15 @@
...
+101: "duet::desktop_numbers_that_matter"
```
fbertsch, line 15:
```suggestion
66: "duet::desktop_numbers_that_matter"
```

src/sync_dashboards/main.py
jasonthomas: This should be fine to do.

PR #82, opened 2021-06-13
iOS documentation

ksiegler1:
Dashboard documentation for iOS for initial round of feedback.

PR #83, opened 2021-06-13
Android dashboard documentation

ksiegler1:
Initial documentation for mobile Android products. I made a mistake in my workflow in an earlier PR so this may look like a dupe of that.


acmiyaguchi (approved):
The summary is succinct and explains the important inconsistencies. Should we merge this one in instead of the ios one, since they look to be duplicates?

scholtzan (approved):
This looks pretty much identical to https://github.com/mozilla/looker-spoke-default/pull/82. So I agree, we could merge this one and close the other PR.

PR #87, opened 2021-06-22
Include generated explores for iOS

fbertsch:


firefox_ios/firefox_ios.model.lkml
```diff
@@ -47,7 +48,7 @@ explore: counters {
   }
 }
 
-explore: client_counts {
+explore: +client_counts {
```
wlach, line 51: What does this do?

wlach (approved):

firefox_ios/firefox_ios.model.lkml
fbertsch: This extends the existing explore we're importing from `looker-hub`. Before it was a new explore, not it's updating the fields of the existing explore.

PR #90, opened 2021-07-06
Don't allow building of new forecasts (temporary fix)

fbertsch:
This lets the KPI dashboard load, by not requiring that the PDT rebuild. I'm going to work with Looker support on figuring out why this change in behavior happened.


irrationalagent (approved):
LGTM, I verified that these changes allow the corporate KPI dash to be displayed again.

PR #91, opened 2021-07-13
Remove old firefox-desktop views

irrationalagent:
This removes the old code that powered V0 of the KPI dashboad (that code now lives in its own explore).

Once we merge this, I'll open a different PR that adds some actual content to the (now hidden) Firefox Desktop Explore. 

@fbertsch OK to merge this and then add the firefox-desktop model to that list we talked about? I don't think this will expose anything in the `explore` menu but it will allow me to work off a branch without having to re-do this in dev.


fbertsch (approved):
Yup, let's do it.

PR #94, opened 2021-07-15
Mobile KR YoY updates

ethompso28:
- Added YoY variance for all 4 mobile products


irrationalagent (approved):
Thanks for adding these in

PR #95, opened 2021-07-20
make dau moving average un-hidden

irrationalagent:
this just unhides this measure so I can use it another dashboard (and have it visible by others)


ethompso28 (approved):
r+

PR #97, opened 2021-07-22
Add views for Desktop H2 Forecast / KRs

irrationalagent:
I built a dashboard based on these views, which can be viewed [here](https://mozilla.cloud.looker.com/dashboards-next/146) after switching to the `h2-kps` branch on spoke default. 

For now, this is just intended to facilitate a the above dashboard that product can use to pitch an update forecast / stretch goal for CDOU. This may not be the final H2 forecast, but we can swap in another one (by changing the underlying table) without touching anything in the code here. 

Briefly: 

`h2_desktop_forecast.view` contains an updated desktop forecast for H2
`h2_desktop_actuals.view` is a simplification of what's already in `firefox_desktop_usage_2021` but without the support for per-segment forecasts. This may not have been strictly necessary, but (1) it makes things load much faster and (2) illustrates how much simpler things could be if we decided to pitch per-segment forecasts for our KPI data model.

We may decide to pull this out or hide it or change it if we decide to refactor the KPIs - I've tried to make it clear by adding (Preliminary) in the explore view.

### DIFF
--- a/KPI/kpi.model.lkml
+++ b/KPI/kpi.model.lkml
@@ -61,6 +61,7 @@ explore: mobile_usage_fields {
 explore: h2_desktop_actuals {
   label: "H2 Desktop"
   group_label: "KPIs"
+  hidden: no
   from: h2_desktop_actuals
   join: h2_desktop_forecast {
     view_label: "H2 Forecast"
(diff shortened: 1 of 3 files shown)


ethompso28 (approved):
I think this looks ok...will approve. What is the plan for integrating this into the current KR dashboard? will this replace it? if we have both will it cause confusion?

irrationalagent:
@ethompso28 I think that assuming they want to adopt the new forecast and want to change the H2 target level, then I see us modifying the current dashboard so that it contains both year-long numbers and H2 specific numbers, relative to the new forecast. 

To be conservative, when we make that change to the official board, we can fork off an "original" version to keep for reference (and clearly mark it, etc). 

But I didn't want to change anything in the current 2021 views yet until that's official, that's why I kept this completely separate so it can be completely pulled out without having to revert any changes to the original KPI views and dashboard.

PR #98, opened 2021-07-23
Group vpn attribution fields

relud:
and hide id and duplicate country field

### DIFF
--- a/mozilla_vpn/views/subscriptions.view.lkml
+++ b/mozilla_vpn/views/subscriptions.view.lkml
@@ -22,6 +22,14 @@ view: +subscriptions {
     primary_key: yes
   }
 
+  dimension: user_id {
+    hidden: yes
+  }
+
+  dimension: country {
+    hidden: yes
+  }
+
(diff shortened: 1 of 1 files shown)


mozilla_vpn/views/subscriptions.view.lkml
```diff
@@ -22,6 +22,14 @@ view: +subscriptions {
     primary_key: yes
   }
 
+  dimension: user_id {
+    hidden: yes
+  }
+
+  dimension: country {
+    hidden: yes
```
acmiyaguchi, line 30: Why hide the country?

acmiyaguchi (approved):
Grouping attribution together looks good with question down below. Looks fine otherwise.

mozilla_vpn/views/subscriptions.view.lkml
relud: yeah, we have country and country_name, and we should only be exposing country_name in the explore.

PR #100, opened 2021-07-28
add revised forecast for mobile h2

irrationalagent:
Similar to last week's PR for desktop, this adds a preliminary new forecast for mobile in H2. 

For now, this only powers [this dashboard](https://mozilla.cloud.looker.com/dashboards-next/156), which is not yet official - product will confirm that we are OK with these forecasts / stretch goals.

Once we get the signal that these forecasts will / will not be adopted, I will work on integrating them into the existing KPI looker framework, and potentially re-factor it so those views are simpler (closer to what's in this PR, potentially with the on-the-fly forecasting capacity pulled out).

### DIFF
--- a/KPI/kpi.model.lkml
+++ b/KPI/kpi.model.lkml
@@ -79,3 +79,25 @@ explore: h2_desktop_actuals {
   # sql_always_where: ${date} >= "2021-07-01" ;;
   }
 }
+
+explore: h2_mobile_actuals {
+  label: "H2 Mobile (Preliminary)"
+  group_label: "KPIs"
+  hidden: no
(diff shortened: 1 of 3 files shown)


irrationalagent:
@ethompso28 I'm waiting to get some clarification on the target level for mobile, I think its still OK to review for now but let's hold on merging

irrationalagent:
OK I got feedback on the target level, this is ready now

ethompso28 (approved):
i think this is fine, assuming the content will migrate over to the main KR dashboards when it's ready.

irrationalagent:
Yea it will, I will do that in a separate PR

PR #102, opened 2021-07-29
Update FxA Views and Explores

fbertsch:
- Include first_seen_date for Growth Accounting
- Add submission field for events analysis (instead of timestamp field)

I think it makes sense to keep these explore extensions in the model file, but open to moving them to their own explores if we think it's a better path.

### DIFF
--- a/firefox_accounts/firefox_accounts.model.lkml
+++ b/firefox_accounts/firefox_accounts.model.lkml
@@ -1,7 +1,23 @@
 connection: "telemetry"
 label: "Firefox Accounts"
+include: "//looker-hub/firefox_accounts/views/fxa_first_seen_table.view.lkml"
 include: "//looker-hub/firefox_accounts/explores/*"
-include: "//looker-hub/firefox_accounts/dashboards/*"
 include: "views/*"
-include: "explores/*"
-include: "dashboards/*"
+
+explore: +growth_accounting {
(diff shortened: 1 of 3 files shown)


firefox_accounts/firefox_accounts.model.lkml
```diff
@@ -1,7 +1,23 @@
 connection: "telemetry"
 label: "Firefox Accounts"
+include: "//looker-hub/firefox_accounts/views/fxa_first_seen_table.view.lkml"
 include: "//looker-hub/firefox_accounts/explores/*"
-include: "//looker-hub/firefox_accounts/dashboards/*"
 include: "views/*"
```
irrationalagent, line 5: This made my head spin a bit but I think I get it now.

So having this include statement basically forces the extensions of the explores below to use the extended versions of the views (defined in e.g. `firefox_accounts/views/events.view.lkml`). This is what essentially overrides the implicit reliance of the imported explores on the original (un-extended) version of the views.

irrationalagent (approved):

firefox_accounts/firefox_accounts.model.lkml
fbertsch: Right, you got it. Some explanation here: https://docs.looker.com/data-modeling/learning-lookml/refinements#refinements_are_applied_in_order.

PR #106, opened 2021-08-03
Add distinct ids to headers on Corporate KPI dashboard

jklukas:
The idea here is that we could provide a link on https://mana.mozilla.org/wiki/pages/viewpage.action?pageId=130932724 that would jump directly to the mobile part of this dashboard

The link would look like: https://mozilla.cloud.looker.com/dashboards-next/101?Date=after%202021%2F01%2F01#mobile_dou

I was not able to get this change to render for me in Looker, and I'm not sure why.

### DIFF
--- a/KPI/dashboards/corporate_kpis.dashboard.lookml
+++ b/KPI/dashboards/corporate_kpis.dashboard.lookml
@@ -146,7 +146,7 @@
     title_text: ''
     subtitle_text: ''
     body_text: |-
-      <h1 style="margin-top:0px; padding: 5px; border-bottom: solid 1px #412399; height: 50px; color: #412399; text-align: left;" id="dou">Desktop Days of Use and DAU</h1>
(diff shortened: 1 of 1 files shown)


irrationalagent (approved):

irrationalagent:
Maybe it didn't work because the script that links the lookml code to the dashboard needs to run first, but im not sure.

jklukas:
This is now applied, but sadly it only semi-works. When I follow a [link with anchor](https://mozilla.cloud.looker.com/dashboards-next/101?Date=after%202021%2F01%2F01#mobile_dou) it still renders the page at the top. If I then go to the location bar in Firefox and hit enter again, it jumps me to the correct header.

I'm guessing this is because Looker builds the content of the page dynamically via JS, so the full document is not available when the page is initially loaded and Firefox is applying the anchor from the URL. At least I learned some things about the workflow for making a PR to this repo.

PR #108, opened 2021-08-04
sort mobile quarterly view correctly in corporate dashboard

irrationalagent:
in the quarterly table, the quarter numbers were out of order in the mobile table. this fixes that


ethompso28 (approved):
Oddly, I had made an H1 custom field for Dan's QBR slide, but I must not have copied it over to the LookML. For now since the QBR is over I don't think this matters, but at the end of Q3 we'll probably be asked to do a YTD table for cDOU

PR #111, opened 2021-08-11
refactor desktop KRs

irrationalagent:
Something like this is intended to eventually replace the KPI explore / views, but I'd like to shop it around and ensure that I can replicate the original KPI dashboard first before I pull the trigger and pull the old stuff out. So for now, this would live in parallel. 

Here are the major changes:

1. There is no more automatic forecasting for segments. We only load forecasts for the aggregate day-level time series. 
2. There is now separate views for the original forecasts (that the KPI targets were set against) and updated forecasts. The latter are derived from a table that I push to manually for now, but I will be working on an automated solution in the coming month. 
3. The query in the `desktop_dau_actuals` view still churns through a decent amount of data, but I THINK I've made it cache and increment on a daily basis - Frank I'm specifically looking for feedback on whether I'm doing this right. Idea is that we only have to run the most recent submission_date, but I'm not sure if that's actually what I'm doing here.
4. For the updated forecasts, I'd like to only filter for the most recent forecast by default but allow people to display older forecasts if desired. Right now I'm doing this by ranking the forecast run dates by descending (so the most recent is always 1) and always filtering in the explore definition for `1`. But I'd like a better solution - one where we automatically filter for the most recent `forecast_run_date`, which would be more understandable

This is a lot, so happy to jump on a call to clarify things if that's faster.


irrationalagent:
@jklukas OK I think I'm done fussing with this now

jklukas (approved):
I'm fine with this getting merged, considering we're marking all of this as WIP.

I think you're right that there's no way to set up a filter to default to a max value, so your workaround of having a recency index is probably the best we can do.

PR #112, opened 2021-08-11
Update vpn retention fields

relud:


mozilla_vpn/views/subscriptions.view.lkml
```diff
@@ -113,35 +113,30 @@ view: +subscriptions {
   }
 
   dimension: billing_grace_period {
-    type: number
```
acmiyaguchi, line 116: Wouldn't the type still be relevant here?

acmiyaguchi (approved):

mozilla_vpn/views/subscriptions.view.lkml
relud: the dimension is actually an `INTERVAL`, which looker doesn't directly support

PR #113, opened 2021-08-12
refactor mobile KR explore and views

irrationalagent:
This is very similar to #111 but for mobile.

This does not modify or pull out any of the existing KPI views, explores, or dashboards for now it just adds the re-factor side-by-side. 

1. This creates a "Mobile Days of Use KRs (Work In Progress)" explore which has the DAU actuals, original January Baseline forecast, and Updated Forecast (each of these live in their own views but are joined together)
2. There is also a mobile_kr_refactor lookml MVP dashboard that can be used for tracking mobile's H2 CDOU KR.

There are also some tweaks to the Desktop refactor here that I snuck in.


irrationalagent:
@ethompso28 since jeff is out today... this contains a fix for desktop which I need to merge before we share with stakeholders

ethompso28 (approved):
r+ for the MVP for stakeholders

PR #118, opened 2021-08-19
update desktop kr dashboard

irrationalagent:
Right now, for the H2 desktop KRs we are pointing folks at this dashboard https://mozilla.cloud.looker.com/dashboards-next/181

That dash was forked off an earlier version which was defined in the `desktop_kr_refactor.dashboard.lookml` file.

This PR updates that file to be consistent with the dashboard at the first link above. 

@magorlick this addresses why those two dashboards look different.


scholtzan (approved):
👍

PR #125, opened 2021-08-31
Add countries view to shared namespace

scholtzan:
This adds a new `shared` namespace to Looker which will contain views that can be used across all namespaces.

Using the `country` view in some dashboard is as simple as adding:

```
include: "../shared/views/countries.view.lkml"
```


focus_android/focus_android.model.lkml
```diff
@@ -1,7 +1,7 @@
...
 include: "//looker-hub/focus_android/explores/*"
-include: "//looker-hub/focus_android/dashboards/*"
-include: "views/*"
-include: "explores/*"
-include: "dashboards/*"
\ No newline at end of file
+# include: "//looker-hub/focus_android/dashboards/*"
+# include: "views/*"
+# include: "explores/*"
+# include: "dashboards/*"
```
scholtzan, line 7: Commenting those out as Looker is showing a warning since these folders do not contain any files

shared/views/countries.view.lkml
```diff
@@ -0,0 +1,26 @@
...
+    description: "ISO 3166 alpha-2 country code"
+  }
+
+  dimension: tier {
+    type: string
+    sql:
+      CASE
+      WHEN
+        ${TABLE}.code IN ('US', 'CA', 'DE', 'FR', 'GB') THEN 'tier-1'
+      ELSE 'non-tier-1'
```
scholtzan, line 22: I'm also open to renaming this to `Tier 1` or something else.

scholtzan:
@jklukas does this match with what you had in mind when trying to create a "Country Details" view?

shared/views/countries.view.lkml
jklukas: I think the goal for this view should be labels that look nice on charts, so I think "Tier 1" is more appropriate than "tier-1".

I've seen lots of redash queries that are labeled for "ROW" (rest of world) as the counterpart to "Tier 1". I do find the ROW term a bit dismissive, which is problematic. "Other" might be good here.

@xluo-ds - You've done some work around defining more granular tiers for different business contexts. Do you see codifying a reusable tier definition here problematic given that work. Any suggestions for terminology here?

shared/views/countries.view.lkml
```diff
@@ -0,0 +1,26 @@
...
+  dimension: name {
+    type: string
+    sql: ${TABLE}.name ;;
+    description: "Official country name per ISO 3166"
+  }
+
+  dimension: code {
+    type: string
+    sql: ${TABLE}.code ;;
+    description: "ISO 3166 alpha-2 country code"
```
jklukas, line 13:
```suggestion
    description: "ISO 3166 alpha-2 country code"
    primary_key: true
```

jklukas (approved):
r+wc

This looks very much in line with what I was thinking. My only concern here is what values we specifically want to enable for tier.

shared/views/countries.view.lkml
xluo-ds: @jklukas  The tier-structure work is still in process, and hopefully in a better shape toward the end of September. In this case, I think we can use either (1) 'Tier 1' and 'non Tier 1' or (2) 'Tier 1' and 'Other'. However, the specific meaning/understanding of "Tier 1" may vary among different products so it would be great if we make annotation available in Looker dashboard.

shared/views/countries.view.lkml
```diff
@@ -0,0 +1,27 @@
...
+
+  dimension: tier {
+    type: string
+    sql:
+      CASE
+      WHEN
+        ${TABLE}.code IN ('US', 'CA', 'DE', 'FR', 'GB') THEN 'Tier 1'
+      ELSE 'Other'
+      END ;;
+    description: "Country tier. The specific meaning of may Tier 1 vary across different products."
```
jklukas, line 25:
```suggestion
    description: "Country tier, as used in relation to browser KPIs. The specific meaning of "Tier 1" may vary across different products."
```

PR #137, opened 2021-09-13
Dashboard layout changes and content updates

ksiegler1:
Updates to the layout of the iOS dashboard (removing an unnecessary step) along with content changes to Android version (removing event installs and adding first time installs)

### DIFF
--- a/duet/dashboards/mobile_android_country.dashboard.lookml
+++ b/duet/dashboards/mobile_android_country.dashboard.lookml
@@ -10,7 +10,7 @@
     body_text: |-
       The purpose of this dashboard is to provide insight into the mobile acquisition funnel for some of our Android products, namely Fenix and Focus. It does not behave like a normal funnel due to the fact that the aggregated data originates from multiple sources, and it's important to understand each source and the caveats that may come with the data.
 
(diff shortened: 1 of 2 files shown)


duet/dashboards/mobile_android_country.dashboard.lookml
```diff
@@ -249,31 +248,33 @@
...
     - model: duet
       explore: mobile_android_country
       type: table
-      fields: [mobile_android_country.join_field, mobile_android_country.event_installs]
+      fields: [mobile_android_country.join_field, mobile_android_country.first_time_installs]
       fill_fields: [mobile_android_country.join_field]
       filters:
         mobile_android_country.period_offset: '1'
-      sorts: [mobile_android_country.event_installs desc]
+        country_buckets.bucket: ''
```
ksiegler1, line 272: @acmiyaguchi I noticed this filter is a new addition. In the UI when I was making changes it says this filter is required so I didn't remove it. When I validated the LookML everything checked out and when I filtered on countries it seems to work okay. Just wanted to call it out in case it is unnecessary or needs to be removed.

acmiyaguchi: It's probably fine, interesting that it got changed though.

acmiyaguchi (approved):
I took a look at this branch in Looker, the changes look good to me.

PR #138, opened 2021-09-14

dataops-looker:


jklukas:
cc @Dexterp37 Do we want Looker content for mlhackweek?

scholtzan:
`mlhackweek_search` get [ignored by lookml-generator](https://github.com/mozilla/lookml-generator/commit/408dc7b7031a5230dee35f4be1641c7d4369cb4b). I forgot to publish a new release though after adding it to the disallow list, that's why the PR got opened here and the Airflow tasks are failing.

Last time I checked, there were no views or explores for `mlhackweek_search`, so merging this would result in Looker being broken.

PR #139, opened 2021-09-14
Fix subscription events sum(delta) calculation

relud:
subscription events join now references `${metadata.last_modified_date}`, which caused a conflict between the many_to_one and one_to_many joins and broke aggregation. Using a one_to_one join here makes Looker understand that the `metadata` join does not impact aggregation. This only works because the `metadata` join does not contain any aggregates.

cc @wwyc

### DIFF
--- a/mozilla_vpn/explores/subscriptions.explore.lkml
+++ b/mozilla_vpn/explores/subscriptions.explore.lkml
@@ -8,6 +8,9 @@ explore: subscriptions {
     from: information_schema_partitions
     view_label: "Metadata"
     sql_on: ${metadata.table_name} = "all_subscriptions_v1" AND ${metadata.partition_id} IS NULL;;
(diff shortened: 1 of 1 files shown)


acmiyaguchi (approved):

mozilla_vpn/explores/subscriptions.explore.lkml
```diff
@@ -8,7 +8,7 @@ explore: subscriptions {
     from: information_schema_partitions
     view_label: "Metadata"
     sql_on: ${metadata.table_name} = "all_subscriptions_v1" AND ${metadata.partition_id} IS NULL;;
-    relationship: many_to_one
+    relationship: one_to_one
```
acmiyaguchi, line 11: Maybe leave a comment above the relationship on why this is necessary.

PR #142, opened 2021-09-22
Add VAT rates to ARR calculation for Mozilla VPN

relud:


mozilla_vpn/views/subscriptions.view.lkml
```diff
@@ -180,6 +180,22 @@ view: +subscriptions {
...
+    sql: CASE
+    WHEN
+      ${plan_interval} = "year"
+    THEN
+      1 / ${plan_interval_count}
+    WHEN
+      ${plan_interval} = "month"
+    THEN
+      12 / ${plan_interval_count}
+    END * ${plan_amount} * (1 - IFNULL(${vat_rates.vat}, 0)) / 100;;
```
relud, line 197: This can't be done in a custom measure, because custom measures are not allowed to use fields from multiple views, and can't use sum_distinct.

wwyc: Thanks for the note.  This is really helpful to know.

wwyc (approved):
Thanks for doing this! It looks good 👍

PR #152, opened 2021-10-05
create views and dashboard for focus relaunch

irrationalagent:
this is in service of https://mozilla-hub.atlassian.net/browse/DS-1735 and https://mozilla-hub.atlassian.net/browse/DS-1698#icft=DS-1698 I've included a hard-coded dashboard that you can preview by switching to the (poorly named) mobile_kr_update branch

anna, if you wouldn't mind looking over the lookml, I can have marissa review the actual dashboard. if you're swamped, feel free to pass to someone else.

note I had to use the legacy telemetry here, since GLEAN is too new on these apps to catch everyone.

### DIFF
--- a/focus_android/focus_android.model.lkml
+++ b/focus_android/focus_android.model.lkml
@@ -1,3 +1,17 @@
 connection: "telemetry"
 label: "Firefox Focus for Android"
 include: "//looker-hub/focus_android/explores/*"
+include: "views/*"
+
+explore: usage {
+  always_filter: {
+    filters: [
+      submission_date: "26 weeks",
+    ]
+  }
+  sql_always_where: submission_date >= "2020-01-01" ;;
+}
+
+explore: focus_launch_new_profile_forecasts {
+  sql_always_where: date >= "2020-01-01" ;;
+}
--- a/focus_ios/focus_ios.model.lkml
+++ b/focus_ios/focus_ios.model.lkml
@@ -2,6 +2,15 @@ connection: "telemetry"
 label: "Firefox Focus for iOS"
 include: "//looker-hub/focus_ios/explores/*"
 # include: "//looker-hub/focus_ios/dashboards/*"
-# include: "views/*"
+include: "views/*"
 # include: "explores/*"
 # include: "dashboards/*"
+
+explore: usage {
+  always_filter: {
+    filters: [
+      submission_date: "26 weeks",
+    ]
+  }
+  sql_always_where: submission_date >= "2020-01-01" ;;
+}
--- a/focus_ios/views/usage.view.lkml
+++ b/focus_ios/views/usage.view.lkml
@@ -1,0 +1,202 @@
+view: usage {
+  derived_table: {
+    sql_trigger_value: SELECT CURRENT_DATE() ;;
+    sql:
+      SELECT
+        submission_date,
+        client_id,
+        days_created_profile_bits,
+        days_seen_bits,
+        days_since_created_profile,
+        days_since_seen,
+        first_seen_date,
+        country,
+        locale
+      FROM `telemetry.nondesktop_clients_last_seen`
+      WHERE product = 'Focus iOS'
+      AND submission_date >= DATE(2019,1,1);;
+  }
+
+  dimension_group: submission {
+    type: time
+    timeframes: [
+      raw,
+      date,
+      week,
+      month,
+      quarter,
+      year,
+    ]
+    convert_tz: no
+    datatype: date
+    sql: ${TABLE}.submission_date ;;
+  }
+
+  dimension: is_new_profile {
+    type: yesno
+    sql: ${days_since_created_profile} = 0 ;;
+  }
+
+  dimension: days_since_created_profile {
+    type: number
+    sql: ${TABLE}.days_since_created_profile ;;
+  }
+
+  dimension: days_seen_bits {
+    type: number
+    sql: ${TABLE}.days_seen_bits ;;
+    hidden: yes
+  }
+
+  dimension: country {
+    type: string
+    sql: ${TABLE}.country ;;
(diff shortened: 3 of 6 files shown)


scholtzan:
Sure, I can take a closer look tomorrow.

focus_ios/views/usage.view.lkml
```diff
@@ -0,0 +1,202 @@
...
+        submission_date,
+        client_id,
+        days_created_profile_bits,
+        days_seen_bits,
+        days_since_created_profile,
+        days_since_seen,
+        first_seen_date,
+        country,
+        locale
+      FROM `telemetry.nondesktop_clients_last_seen`
```
scholtzan, line 15: I see this table is also used for focus_android. It might make sense to have the `nondesktop_clients_last_seen` view in looker-hub instead of having it as a derived table here. Then the existing view could simply be extended by additional dimensions without having to manually maintain existing dimensions, like `country` or `submission_date`. I guess we could create a `nondesktop` namespace for this view and related ones.

The filtering could also be done in the explore by adding `sql_always_where`. This way we wouldn't need to use `derived_table` here, which is something we'd generally want to avoid (extending views is easier to maintain than custom written SQL, also changes, like adding fields, can propagate easier).

That being said, I'm also OK with merging this as is. Just things to keep in mind for future improvements.

focus_ios/focus_ios.model.lkml
```diff
@@ -2,6 +2,15 @@ connection: "telemetry"
...
 # include: "explores/*"
 # include: "dashboards/*"
+
+explore: usage {
+  always_filter: {
+    filters: [
+      submission_date: "26 weeks",
+    ]
+  }
+  sql_always_where: submission_date >= "2020-01-01" ;;
```
scholtzan, line 15: This filter actually overrides the `AND submission_date >= DATE(2019,1,1)` part in the `derived_table` definition.

focus_ios/focus_ios.model.lkml
```diff
@@ -2,6 +2,15 @@ connection: "telemetry"
...
+include: "views/*"
 # include: "explores/*"
 # include: "dashboards/*"
+
+explore: usage {
+  always_filter: {
+    filters: [
+      submission_date: "26 weeks",
+    ]
+  }
```
scholtzan, line 14: I think the dashboard can significantly reduce load times by using [`aggregate_table`](https://docs.looker.com/data-modeling/learning-lookml/aggregate_awareness), especially since it doesn't have any filters that can changed and essentially always runs the same queries. Aggregate tables basically cache previous results and can incrementally add new data.

Defining aggregate tables is actually pretty easy, in fact Looker can generate the lookml for you. If you open an explore and click on the cog icon in the right corner, there is a "Get LookML" entry which shows the "Aggregate Table" LookML that can get copied into the explore definition.

It's not possible to use aggregate tables for some forecasting cases though.

Looking through the dashboard I think we can add aggregate tables for the following:

```suggestion
  }

  aggregate_table: rollup__new_profile_activated__submission_date {
    query: {
      dimensions: [new_profile_activated, submission_date]
      measures: [user_count_active_28_days]
      filters: [
        usage.new_profile_7days_ago: "Yes",
        usage.submission_date: "after 2021/01/01"
      ]
    }
    
    # Whenever CURRENT_DATE() changes, the query will get re-executed and added to the aggregate table
    # The dashboard will then directly query the aggregate table instead of running the query, which makes loading dashboard tiles much faster
    materialization: {
      sql_trigger_value: SELECT CURRENT_DATE();;
    }
  }
  
  aggregate_table: rollup__active_8_days_active_last_28__submission_date {
    query: {
      dimensions: [active_8_days_active_last_28, submission_date]
      measures: [user_count_active_28_days]
      filters: [usage.submission_date: "after 2021/01/01"]
    }

    materialization: {
      sql_trigger_value: SELECT CURRENT_DATE();;
    }
  }

  aggregate_table: rollup__submission_date {
    query: {
      dimensions: [submission_date]
      measures: [user_count_active_this_day]
      filters: [
        usage.is_new_profile: "Yes",
        usage.submission_date: "after 2021/01/01"
      ]
    }
    
    materialization: {
      sql_trigger_value: SELECT CURRENT_DATE();;
    }
  }

```

focus_android/focus_android.model.lkml
```diff
@@ -1,3 +1,17 @@
...
+include: "views/*"
+
+explore: usage {
+  always_filter: {
+    filters: [
+      submission_date: "26 weeks",
+    ]
+  }
+  sql_always_where: submission_date >= "2020-01-01" ;;
+}
```
scholtzan, line 13:
```suggestion
 
  aggregate_table: rollup__new_profile_activated__submission_date {
    query: {
      dimensions: [new_profile_activated, submission_date]
      measures: [user_count_active_28_days]
      filters: [
        usage.new_profile_7days_ago: "Yes",
        usage.submission_date: "after 2021/01/01"
      ]
    }

    materialization: {
      sql_trigger_value: SELECT CURRENT_DATE();;
    }
  }

  aggregate_table: rollup__active_8_days_active_last_28__submission_date {
    query: {
      dimensions: [active_8_days_active_last_28, submission_date]
      measures: [user_count_active_28_days]
      filters: [usage.submission_date: "after 2021/01/01"]
    }

    materialization: {
      sql_trigger_value: SELECT CURRENT_DATE();;
    }
  }

  aggregate_table: rollup__submission_date {
    query: {
      dimensions: [submission_date]
      measures: [user_count_active_this_day]
      filters: [
        usage.is_new_profile: "Yes",
        usage.submission_date: "after 2021/01/01"
      ]
    }

    materialization: {
      sql_trigger_value: SELECT CURRENT_DATE();;
    }
  }
}  
```

PR #185, opened 2021-11-15
Add Pocket section and modify y axes for percentage graphs

ksiegler1:


relud (approved):
giving this an r+ without situational awareness, so that it doesn't have to wait for @scholtzan to return

PR #186, opened 2021-11-15
Update sql trigger value for VPN SaaSboard - Active Subscriptions

wwyc:
The SELECT CURRENT_DATE() sql trigger is not sufficient for updating the PDT required for active subscriptions views in Looker.  The new sql trigger value included should check for changes in the latest count of active subscriptions.

### DIFF
--- a/mozilla_vpn/explores/subscriptions.explore.lkml
+++ b/mozilla_vpn/explores/subscriptions.explore.lkml
@@ -73,7 +73,14 @@ explore: +subscriptions {
     }
 
     materialization: {
-      sql_trigger_value: SELECT CURRENT_DATE();;
+      sql_trigger_value: SELECT (subscriptions__active) AS subscriptions__active_active_date,COUNT(DISTINCT subscriptions.subscription_id ) AS subscriptions_count
+      FROM `mozdata.mozilla_vpn.all_subscriptions`  AS subscriptions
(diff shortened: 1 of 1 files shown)


mozilla_vpn/explores/subscriptions.explore.lkml
```diff
@@ -73,7 +73,14 @@ explore: +subscriptions {
...
     materialization: {
-      sql_trigger_value: SELECT CURRENT_DATE();;
+      sql_trigger_value: SELECT (subscriptions__active) AS subscriptions__active_active_date,COUNT(DISTINCT subscriptions.subscription_id ) AS subscriptions_count
+      FROM `mozdata.mozilla_vpn.all_subscriptions`  AS subscriptions
+      LEFT JOIN UNNEST((GENERATE_DATE_ARRAY((DATE(subscriptions.subscription_start_date )), (DATE(subscriptions.end_date )) - 1))) AS subscriptions__active
+      GROUP BY
+          1
+      ORDER BY
+          1 DESC
+      LIMIT 1;;
```
relud, line 83: I think a more correct trigger here is the last modified timestamp for the all subscriptions table:
```suggestion
      sql_trigger_value: SELECT
        MAX(last_modified_time)
      FROM
        moz-fx-data-shared-prod.mozilla_vpn_derived.INFORMATION_SCHEMA.PARTITIONS
      WHERE
        table_name = "all_subscriptions_v1";;
```

relud (approved):

PR #192, opened 2021-11-17
Bug(LOOK-117): Incorrect install count in "Numbers that matter" for 2021-11-01

kik-kik:
# Bug(LOOK-117): Incorrect install count in "Numbers that matter" for 2021-11-01

Bug ticket: https://mozilla-hub.atlassian.net/browse/LOOK-117

## Description
"Numbers that matter" dashboard incorrectly shows a near-zero install count for 2021-11-01. This is due to the time gap between builds `93.0` and `94.0` being longer than 1 month.

## Change
Grabbing build from the last 6 weeks instead of 1 month to avoid counting issues if longer time between builds.

This is what the graph looks like after the change:
image


wlach (approved):
Thank you!

PR #199, opened 2021-11-19
Add aggregate tables to speed up newtab dashboard

emtwo:
To address [this issue](https://mozilla-hub.atlassian.net/browse/DO-636) of the new tab dashboard being slow to load, I added the recommended aggregate tables from Looker.

I think the dashboard in the New Tab folder might be a copy of the one in `looker-spoke-private`? But I think these changes should still work then.

Tagging Megan since it's your dashboard and Anna in case this isn't set up in the right place.

### DIFF
--- a/search/search.model.lkml
+++ b/search/search.model.lkml
@@ -17,3 +17,49 @@ explore: +mobile_search_counts {
     SAP, and organic searches are those that occur directly on a search webpage (e.g. www.google.com).
     Warning: Firefox iOS is not able to implement all metrics, like ad clicks."
 }
+
+explore: +desktop_search_counts {
+  aggregate_table: rollup__search_clients_engines_sources_daily_submission_date__0 {
+    query: {
+      dimensions: [search_clients_engines_sources_daily.submission_date]
+      measures: [search_clients_engines_sources_daily.total_searches]
+      filters: [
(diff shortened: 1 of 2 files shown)


search/search.model.lkml
```diff
@@ -17,3 +17,49 @@ explore: +mobile_search_counts {
...
     Warning: Firefox iOS is not able to implement all metrics, like ad clicks."
 }
+
+explore: +desktop_search_counts {
+  aggregate_table: rollup__search_clients_engines_sources_daily_submission_date__0 {
+    query: {
+      dimensions: [search_clients_engines_sources_daily.submission_date]
+      measures: [search_clients_engines_sources_daily.total_searches]
+      filters: [
+        search_clients_engines_sources_daily.source: "newtab",
```
scholtzan, line 27: Without having much context on the data here, it looks like the other filters on `source` to have wildcards in them. Not sure if that should be the same here?
```suggestion
        search_clients_engines_sources_daily.source: "%newtab%",
```

scholtzan (approved):
The setup looks good to me

search/search.model.lkml
emtwo: Good observation. I just checked and this rollup uses the filter with source specifically equal to `newtab` so this is intentional.

PR #214, opened 2021-12-09

dataops-looker:


scholtzan (changes requested):
disallowed in https://github.com/mozilla/lookml-generator/pull/325

PR #216, opened 2021-12-17
Vpn acquistion funnel view

YeonjooSmith:
Need to update the view files to share the dashboard MVP for feedback.


mozilla_vpn/views/test1.view.lkml
```diff
@@ -0,0 +1,274 @@
+view: test1 {
```
scholtzan, line 1: I'm guessing this will just be temporary, otherwise this view should probably have a more descriptive name

scholtzan (approved):

mozilla_vpn/views/test1.view.lkml
YeonjooSmith: @scholtzan  I removed test1 view and explore files now and pushed.

scholtzan: oh, you'll need to open another PR

PR #228, opened 2022-01-03
DVPN-77 - Migrate SaaSboards to new aggregate tables

relud:
[DVPN-77](https://mozilla-hub.atlassian.net/browse/DVPN-77)

this leaves out the revenue saasboard because it has some issues that still need to be worked out


wwyc (approved):
I think the [PDTs](https://github.com/mozilla/looker-spoke-default/blob/94e16cd07a433641ca37071b4c573123f6ae5f94/mozilla_vpn/explores/subscriptions.explore.lkml#L63) would need to be adjusted as well?  I can make those changes after you merge.  Thanks!

PR #242, opened 2022-01-25
Detect new name for Apple Store provider

relud:
for https://github.com/mozilla/bigquery-etl/pull/2672

### DIFF
--- a/mozilla_vpn/views/subscriptions.view.lkml
+++ b/mozilla_vpn/views/subscriptions.view.lkml
@@ -112,7 +112,7 @@ view: +subscriptions {
   dimension: cancel_reason {
     type: string
     sql: CASE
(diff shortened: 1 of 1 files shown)


mozilla_vpn/views/subscriptions.view.lkml
```diff
@@ -112,7 +112,7 @@ view: +subscriptions {
   dimension: cancel_reason {
     type: string
     sql: CASE
-      WHEN ${provider} LIKE "Apple Store IAP" THEN "Cancelled by IAP"
```
wwyc, line 115: Should this still be named as "Cancelled by IAP"?

relud: we could rename it to `Cancelled by Apple`, but that will break dashboards

wwyc (approved):

PR #243, opened 2022-01-26
Add exchange rate view to vpn subscriptions

wwyc:
Adding exchange rate view to support ARR verification


mozilla_vpn/views/exchange_rates.view.lkml
```diff
@@ -0,0 +1,41 @@
+include: "//looker-hub/mozilla_vpn/views/exchange_rates_table.view"
```
relud, line 1: this view file is unnecessary, as it doesn't change any of the values for `exchange_rates_table`

relud (changes requested):

mozilla_vpn/views/exchange_rates.view.lkml
wwyc: Thanks!  I have removed the view.

relud (approved):

PR #245, opened 2022-01-26
Fenix monitoring migration

Marlene-M-Hirose:
fenix monitoring view and explore for database sizes comparisons


monitoring/views/database_sizes_mb.view.lkml
```diff
@@ -0,0 +1,58 @@
+# If necessary, uncomment the line below to include explore_source.
+include: "//looker-hub/fenix/explores/metrics.explore.lkml"
```
scholtzan, line 2: instead of extending the metrics explore, it would have been a little more straightforward to extend the metrics view and define `database_size_sum_mb` and `database_size_tiers` as dimensions.

monitoring/views/database_sizes_mb.view.lkml
```diff
@@ -0,0 +1,58 @@
...
+        value: "mozdata.fenix.metrics"
+      }
+      filters: {
+        field: metrics.submission_date
+        value: "28 days"
+      }
+      filters: {
+        field: metrics.metrics__memory_distribution__glean_database_size__sum
+        value: "NOT NULL"
+      }
```
scholtzan, line 27: generally for filters it's best practice to define filters on an explore level. Having them inside of the view definition hides some information from the user (e.g. why does the data only show me the last 28 days?)

monitoring/explores/database_sizes_mb.explore.lkml
```diff
@@ -0,0 +1,5 @@
+include: "../views//database_sizes_mb.view.lkml"
+
+ explore: database_sizes_mb {
+   view_name: database_sizes_mb
```
scholtzan, line 4: Since we only use this explore for some internal analysis, we can set it to `hidden`. This means the explore won't be explicitly listed in the "Explore"s menu and reduces clutter. 

```suggestion
   view_name: database_sizes_mb
   hidden: yes
```

scholtzan (approved):
I left some suggestions, but I'm fine with merging this as is since it's anyway just for internal use.

monitoring/views/database_sizes_mb.view.lkml
Marlene-M-Hirose: I was thinking about this and I'll make the updates

monitoring/views/database_sizes_mb.view.lkml
Marlene-M-Hirose: good to know. will keep that in mind

PR #255, opened 2022-02-05
Add additional dimensions for usage view 

birdsarah:
(and re-order table construction dau, wau, mau)


scholtzan (approved):
Looking good 👍

PR #257, opened 2022-02-05
Replace usage with usage_mac and usage_wac

birdsarah:
I wanted to be able to dis-aggregate MAC/WAC across various dimensions. 

I did not find a way to do this where mac and wac could be included in the same table.

We don't need a table for DAC anyway as you can make it directly from main.

Update: switched vocab to be more accurate and consistent - these are views of Monthly Active Clients (not users)

### DIFF
--- a/mozilla_vpn/explores/usage_mau.explore.lkml
+++ b/mozilla_vpn/explores/usage_mau.explore.lkml
@@ -1,0 +1,6 @@
+include: "../views/usage_mau.view"
+
+explore: usage_mau {
+  description: "VPN Monthly Active Users (based on Glean telemetry)"
+
+}
--- a/mozilla_vpn/explores/usage_mau.explore.lkml
+++ b/mozilla_vpn/explores/usage_mau.explore.lkml
@@ -2,4 +2,5 @@ include: "../views/usage_mau.view"
 
 explore: usage_mau {
   description: "VPN Monthly Active Users (based on Glean telemetry)"
+  label: "Monthly Usage"
 }
--- a/mozilla_vpn/explores/usage_mau.explore.lkml
+++ b/mozilla_vpn/explores/usage_mau.explore.lkml
@@ -1,6 +1,0 @@
-include: "../views/usage_mau.view"
-
-explore: usage_mau {
-  description: "VPN Monthly Active Users (based on Glean telemetry)"
-  label: "Monthly Usage"
-}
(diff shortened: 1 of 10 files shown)


birdsarah:
Thanks to some tips from @scholtzan I've worked on this locally and verified in my two new explores that things are working as expected

Screenshot from 2022-02-04 20-52-15

Screenshot from 2022-02-04 20-53-23

mozilla_vpn/explores/usage_mau.explore.lkml
```diff
@@ -0,0 +1,5 @@
+include: "../views/usage_mau.view"
+
+explore: usage_mau {
+  description: "VPN Monthly Active Users (based on Glean telemetry)"
```
scholtzan, line 4: You can also give it a custom label (MAU, WAU, DAU might not be clear to everyone maybe). The label is used in the "Explores" list.
```suggestion
  description: "VPN Monthly Active Users (based on Glean telemetry)"
  label: "Monthly Usage"
```

mozilla_vpn/explores/usage_wau.explore.lkml
```diff
@@ -0,0 +1,5 @@
+include: "../views/usage_wau.view"
+
+explore: usage_wau {
+  description: "VPN Weekly Active Users (based on Glean telemetry)"
```
scholtzan, line 4:
```suggestion
  description: "VPN Weekly Active Users (based on Glean telemetry)"
  label: "Weekly Usage"
```

mozilla_vpn/views/usage.view.lkml
```diff
@@ -1,116 +0,0 @@
...
-        glean_mau AS mau,
-        glean_wau AS wau
-      FROM
-        dau
-        LEFT JOIN
-        wau
-        USING (date)
-        LEFT JOIN
-        mau
-        USING (date);;
```
scholtzan, line 74: I do think it would be possible to keep everything in one view, but I'm fine with splitting it up if that makes things easier
```suggestion
    SELECT
        date,
        os,
        app_version,
        glean_dau AS dau,
        glean_mau AS mau,
        glean_wau AS wau
      FROM
        dau
        LEFT JOIN
        wau
        USING (date, os, app_version)
        LEFT JOIN
        mau
        USING (date, os, app_version);;
```

scholtzan (approved):

mozilla_vpn/views/usage.view.lkml
birdsarah: I already tried that and it didn't work for me. 

See this earlier commit in the PR https://github.com/mozilla/looker-spoke-default/pull/257/commits/8bbd64f21e54a32f07b2887096c9a32cf072b1e5

birdsarah:
As I don't have write access I couldn't just accept your suggestions so I pushed a new commit. Thanks for the review. This should be good to merge now.

PR #260, opened 2022-02-09
Added dimensions to support renewal analysis

wwyc:
Three dimensions were ended:
- cancel_at_period_end:  Indicates if auto-renewal was cancelled and subscription is to be cancelled at period end
- is_ended:  Indicates if subscription has ended
- plan_interval_type:  Indicates the plan interval type (1 year, 6 month, 1 month, etc)

### DIFF
--- a/mozilla_vpn/views/subscriptions.view.lkml
+++ b/mozilla_vpn/views/subscriptions.view.lkml
@@ -63,7 +63,7 @@ view: +subscriptions {
   dimension: plan_interval_type {
     description: "Indicates the plan interval type (1 year, 6 month, 1 month, etc)"
     type: string
-    sql: CONCAT(CAST(${TABLE}.plan_interval_count AS STRING),"_",  ${TABLE}.plan_interval);;
+    sql: CONCAT(${plan_interval_count},"_",  ${TABLE}.plan_interval);;
   }
 
   dimension: normalized_source {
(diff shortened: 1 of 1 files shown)


mozilla_vpn/views/subscriptions.view.lkml
```diff
@@ -52,6 +54,18 @@ view: +subscriptions {
...
+  dimension: is_ended {
+    description: "Indicates if subscription has ended"
+    type: yesno
+    sql: ${TABLE}.ended_at IS NOT NULL;;
+  }
+
+  dimension: plan_interval_type {
+    description: "Indicates the plan interval type (1 year, 6 month, 1 month, etc)"
+    type: string
+    sql: CONCAT(CAST(${TABLE}.plan_interval_count AS STRING),"_",  ${TABLE}.plan_interval);;
```
relud, line 66: the `CAST` here shouldn't be necessary, and fields should be referenced as `${field}` not `${TABLE}.field`:
```suggestion
    sql: CONCAT(${plan_interval_count}, "_",  ${plan_interval});;
```

mozilla_vpn/views/subscriptions.view.lkml
```diff
@@ -52,6 +54,18 @@ view: +subscriptions {
...
+    }
```
relud, line 67: indent on this line is inconsistent:
```suggestion
  }
```

relud (approved):
r+wc

PR #265, opened 2022-02-14
Schema errors over time

Marlene-M-Hirose:


monitoring/views/grouped_schema_error_counts.view.lkml
```diff
@@ -0,0 +1,23 @@
+include: "//looker-hub/monitoring/views/schema_error_counts.view.lkml"
+
+view: grouped_schema_error_counts {
+    extends: [schema_error_counts]
+
+  sql_table_name:`moz-fx-data-shared-prod`.monitoring.schema_error_counts;;
```
scholtzan, line 6: should we just extend the existing `schema_error_counts` view here? Having separate `schema_error_counts` and `grouped_schema_error_counts` explores might be more confusing since they are _very_ similar.

```suggestion
view: +schema_error_counts {
```

monitoring/views/overall_ping_counts.view.lkml
```diff
@@ -0,0 +1,14 @@
+include: "distinct_docids.view.lkml"
+
+view: overall_ping_counts {
+  extends: [distinct_docids]
```
scholtzan, line 4: Same here, extending the existing view is maybe better than creating a new one.
```suggestion
view: +distinct_docids {
```

monitoring/views/grouped_schema_error_counts.view.lkml
Marlene-M-Hirose: ah, this works much better. Thanks!

monitoring/explores/overall_ping_counts.explore.lkml
```diff
@@ -0,0 +1,6 @@
+include: "../views//overall_ping_counts.view.lkml"
+
+explore: overall_ping_counts {
+  view_name: overall_ping_counts
+  hidden:  yes
+}
```
scholtzan, line 6: the we can remove the two explores here and it should be good to go

Marlene-M-Hirose: would I have to rebuild the merged explore?

scholtzan: yes

scholtzan (approved):
🚢

PR #268, opened 2022-02-14
Fix distinct_doc_ids join

scholtzan:


monitoring/explores/distinct_docids.explore.lkml
```diff
@@ -20,8 +20,12 @@ explore: distinct_docids {
...
 
   join: non_matching {
-    relationship: one_to_many
-    sql: LEFT JOIN non_matching USING(submission_date) ;;
+    relationship: one_to_one
+    sql: LEFT JOIN non_matching USING(
+      submission_date,
+      doc_type,
+      namespace
+    ) ;;
```
scholtzan, line 28: this should be a 1:1 join so we can determine the offending doc_type/namespace

monitoring/views/distinct_docids.view.lkml
```diff
@@ -61,19 +61,26 @@ view: distinct_docids {
     type: number
   }
 
-
+  measure: n_documents {
+    type:  sum
+    sql: (${TABLE}.stable) ;;
+  }
```
scholtzan, line 67: moving from overall_ping_counts, no need to have two separate view files

Marlene-M-Hirose (approved):

PR #269, opened 2022-02-14
Simplify distinct docids view

scholtzan:


monitoring/views/distinct_docids.view.lkml
```diff
@@ -61,45 +61,15 @@ view: distinct_docids {
     type: number
   }
 
+  measure: non_matching_count {
+    type: number
+    sql: COUNTIF(
+      ${TABLE}.decoded != ${TABLE}.live OR ${TABLE}.decoded != ${TABLE}.stable
+    );;
+  }
+
```
scholtzan, line 70: let's count this stuff in here instead of a separate view

Marlene-M-Hirose (approved):

PR #275, opened 2022-03-02
Kpi 2022 review

scholtzan:


kpi/views/browser_kpis.view.lkml
```diff
@@ -0,0 +1,44 @@
+include: "./unified_metrics.view.lkml"
+
+view: browser_kpis {
+  extends: [unified_metrics]
```
scholtzan, line 4: Extending the view instead of the explore also updates this view when `unified_metrics` is updated (e.g. when new fields are being added)

irrationalagent (approved):
thank you for taking the time to do this. much less brittle than what I was doing

PR #279, opened 2022-03-09
Changes to unified metrics and KPI explores

irrationalagent:
will inline PR comments

### DIFF
--- a/kpi/kpi.model.lkml
+++ b/kpi/kpi.model.lkml
@@ -26,7 +26,7 @@ explore: firefox_desktop_usage_2021 {
     type: cross
     relationship: many_to_one
   }
-  hidden: no
+  hidden: yes
 }
 
 # For suggestions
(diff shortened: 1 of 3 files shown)


kpi/kpi.model.lkml
```diff
@@ -26,7 +26,7 @@ explore: firefox_desktop_usage_2021 {
     type: cross
     relationship: many_to_one
   }
-  hidden: no
```
irrationalagent, line 29: hiding the 2021 KPI related explores from the UI. I'm not removing them because I don't want the old dashboards to break.

kpi/views/browser_kpis.view.lkml
```diff
@@ -15,6 +15,30 @@ view: browser_kpis {
     hidden: yes
   }
 
+  dimension: active_hours_sum {
```
irrationalagent, line 18: I'm hiding these because I want the KPI view to be focused solely on the user count. if people want to dig into these dimensions they can use the unified metrics explore.

kpi/views/browser_kpis.view.lkml
```diff
@@ -35,10 +59,8 @@ view: browser_kpis {
     hidden: yes
   }
 
-  measure: user_count_qualified {
```
irrationalagent, line 38: this is redundant with the unified_metrics.client_qualifies dimension. im removing to avoid confusion.

Marlene-M-Hirose (approved):
Good changes. TY

PR #283, opened 2022-03-16
Labeled total_searches "SAP Searches" RS-268

ANich:
Ref: https://docs.looker.com/reference/field-params/label-for-field


skahmann3 (approved):
Looks great, thank you!

PR #286, opened 2022-03-22
Count measures for urlbar_clients_daily

scholtzan:


rebecca-burwei (approved):
@scholtzan looks great - thank you

PR #291, opened 2022-04-01
create derived view for browser KPIs

irrationalagent:
This creates a looker-native derived table for tracking the browser KPIs. It only contains three fields:

1. submission_date
2. platform
3. DAU / Days of use

The derived table is joined day-by-day onto the forecast / target explore so comparing to forecast / targets does not require a merge.

What was previously the `browser_kpis` explore now is called the `browser_dau` explore and can be used for segmenting, drilling down etc. The important thing here is that we are successfully able to direct people to the dau explore if they want to explore DAU on their own after viewing the KPI dashboard. 

This will temporarily break the temp dashboard for browser CDOU that I have been pointing people to, so I will need to fix that shortly after this being merged (I will stage some charts in some looks that I can copy over so that the breakage shouldn't last too long).

### DIFF
--- a/kpi/kpi.model.lkml
+++ b/kpi/kpi.model.lkml
@@ -104,17 +104,17 @@ explore: recent_mobile_forecast {
 
 explore: browser_kpis {
   label: "2022 Browser KPIs"
-  group_label: "Core Browser Metrics and KPIs"
-
-  always_filter: {
-    filters: [
-      browser_kpis.active_today: "yes"
-    ]
+  group_label: "Official Browser KPIs"
+  description: "For Official Accounting. To drill down into other dimensions (e.g. Country) please use the 'Browsers Daily Active Users' Explore."
+  join: loines_browser_2022_forecasts {
+    type: left_outer
+    sql_on: ${browser_kpis.submission_date} = ${loines_browser_2022_forecasts.date_date} AND ${browser_kpis.platform} = ${loines_browser_2022_forecasts.platform};;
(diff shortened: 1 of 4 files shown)


kpi/kpi.model.lkml
```diff
@@ -123,7 +123,18 @@ explore: +unified_metrics {
   }
 }
 
+explore: browser_dau {
+  label: "Browsers Daily Active Users"
+  group_label: "Core Browser Metrics"
+  description: "Easiest to use for answering most common questions about (Q)CDOU."
+  always_filter: {
+    filters: [
```
irrationalagent, line 131: this will effectively only filter to qualified DAU for desktop (as intended)

kpi/kpi.model.lkml
```diff
@@ -104,17 +104,17 @@ explore: recent_mobile_forecast {
...
-  group_label: "Core Browser Metrics and KPIs"
-
-  always_filter: {
-    filters: [
-      browser_kpis.active_today: "yes"
-    ]
+  group_label: "Official Browser KPIs"
+  description: "For Official Accounting. To drill down into other dimensions (e.g. Country) please use the 'Browsers Daily Active Users' Explore."
+  join: loines_browser_2022_forecasts {
+    type: left_outer
```
irrationalagent, line 110: This is where the join to the forecast happens

Marlene-M-Hirose (approved):

PR #306, opened 2022-04-20
Sort x-axis on VPN churn chart

sean-rose:
On the [VPN SaasBoard - Churn](https://mozilla.cloud.looker.com/dashboards/mozilla_vpn::vpn_saasboard__churn) dashboard the "Churn by Months Since Subscription Start" chart's data is sorted by churn counts, which sometimes results in the x-axis (months since subscription start) being in non-contiguous order when the churn count doesn't drop off in a regular way.  This was noticed by @zerokarmaleft and mentioned in [Slack](https://mozilla.slack.com/archives/G018F8NUQKD/p1650493398177429).

This PR changes the sorting so the "Churn by Months Since Subscription Start" chart's x-axis should be contiguous.

### DIFF
--- a/mozilla_vpn/dashboards/vpn_saasboard_churn.dashboard.lookml
+++ b/mozilla_vpn/dashboards/vpn_saasboard_churn.dashboard.lookml
@@ -56,7 +56,7 @@
     filters:
       subscriptions__retention.months_since_subscription_start: ">0"
       subscriptions__retention.is_cohort_complete: 'Yes'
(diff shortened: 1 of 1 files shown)


sean-rose:
@wwyc I don't have developer access to be able to preview the result of this branch, and I'm also a Looker novice, so if you wouldn't mind please verify this works as intended in developer mode.

wwyc (approved):

PR #309, opened 2022-04-26
Fix dimensions that should be measures in Client Counts

rebecca-burwei:


rebecca-burwei:
@scholtzan could you help me with this PR? I tried to follow the pattern in this [PR](https://github.com/mozilla/looker-spoke-default/pull/286) to fix Looker Dimensions that should be Measures. However, I get the following error:

image

firefox_desktop/firefox_desktop.model.lkml
```diff
@@ -82,3 +82,146 @@ explore: +metrics {
...
   }
 }
+
+view: +clients_daily_table__contextual_services_quicksuggest_block_nonsponsored_bestmatch_sum {
+  dimension: value {
+    hidden: yes
+  }
+
+  measure: count {
+    sql: ${TABLE}.value ;;
```
scholtzan, line 92: You'll need to do some aggregation here
```suggestion
   sql: SUM(${TABLE}.value) ;;
```
For the awesome_bar view the underlying data was already aggregated/grouped but that doesn't seem to be the case here.

scholtzan:
So this should be fixed now.
The reason was the for you `looker-hub` was not up-to-date. I went ahead and pulled in remote changes for you, so the error should be gone now.

firefox_desktop/firefox_desktop.model.lkml
scholtzan: This is also necessary for all the other views

rebecca-burwei:
@scholtzan thanks! I have made those changes. PR is ready for review.

scholtzan (approved):

PR #323, opened 2022-05-13
Feat: adding Mozilla websites metrics and summary views to Looker

kik-kik:
# Feat: adding Mozilla websites metrics and summary views to Looker


websites/views/blogs_summary.view.lkml
```diff
@@ -0,0 +1,10 @@
+include: "//looker-hub/websites/views/blogs_daily_summary.view.lkml"
+
+view: blogs_summary {
+  extends: [blogs_daily_summary]
+
+  filter: date {
```
gkabbz, line 6: I don't think I've seen this in other explores. What does it do?

kik-kik: Since I'm not very familiar with Looker I just grabbed that from another existing view:
https://github.com/mozilla/looker-spoke-default/blob/main/user_journey/views/cohort_analysis/cohort_analysis.view.lkml#L6-L9

In Looker this is what I see:
image

websites/websites.model.lkml
```diff
@@ -1,2 +1,4 @@
 connection: "telemetry"
-label: "Websites"
\ No newline at end of file
+label: "Websites"
+
+include: "/websites/views/*.view.lkml"
```
gkabbz, line 4: You still need to add the explores. For [example ](https://github.com/mozilla/looker-spoke-default/blob/main/firefox_accounts/firefox_accounts.model.lkml)

websites/views/blogs_summary.view.lkml
kik-kik: image


So this seems as a way for defining a filter field. Original example was casting to `timestamp`, looks like that might be needed:
image

websites/websites.model.lkml
gkabbz: The explores are what the end users see in looker and pick from

websites/views/blogs_summary.view.lkml
kik-kik: This is what it looks like when you click on the filter field in Looker UI: 
image

websites/websites.model.lkml
kik-kik: Added it. Now I can see these views when going into `Explore`

kik-kik:
image

Now these are showing correctly, and it appears you can explore the views.

gkabbz (approved):

PR #324, opened 2022-05-13
Clean presentation of active user aggregates

gkabbz:
Converting measures defined as dimensions by lookml-generator back to measures and grouping dimensions


combined_browser_metrics/views/active_users_aggregates.view.lkml
```diff
@@ -109,9 +109,162 @@ dimension: ytd_only {
...
+  }
+
+  measure: monthly_active_users {
+    label: "MAU"
+    type: sum
+    sql:  ${TABLE}.mau ;;
+  }
+
+  measure: new_profile {
+    label: "New Profile"
```
ANich, line 183: Looker's convention is (1) replace `_` with space and (2) capitalize, so some of these like this one will automatically turn into "New Profile" but I think it is better to be explicit 👍

ANich (approved):

PR #345, opened 2022-06-01
Adjusted grace period in current_months_since_cohort_complete

wwyc:
Some views in the VPN SaaSboard retention & churn dashboards rely on current_months_since_cohort_complete filter to hide data from most recent cohort.  Because mobile subscriptions have billing grace period = 0, this is causing those subscriptions to appear before stripe subscriptions that have a 7 days billing grace period.

### DIFF
--- a/mozilla_vpn/views/subscriptions.view.lkml
+++ b/mozilla_vpn/views/subscriptions.view.lkml
@@ -155,7 +155,7 @@ view: +subscriptions {
     sql: mozfun.norm.diff_months(
       start => LAST_DAY(DATE(${subscriptions.subscription_start_raw}), MONTH),
       `end` => DATE(${metadata.last_modified_date}),
-      grace_period => ${subscriptions.billing_grace_period},
+      grace_period => INTERVAL 7 DAY,
       inclusive => FALSE
     );;
   }


mozilla_vpn/views/subscriptions.view.lkml
```diff
@@ -155,7 +155,7 @@ view: +subscriptions {
     sql: mozfun.norm.diff_months(
       start => LAST_DAY(DATE(${subscriptions.subscription_start_raw}), MONTH),
       `end` => DATE(${metadata.last_modified_date}),
-      grace_period => ${subscriptions.billing_grace_period},
+      grace_period => INTERVAL 7 DAY,
```
sean-rose, line 158: Looking at the data the billing grace period for Google Play appears to be upwards of a month (current max I'm seeing is 26 days, [the ETL uses 1 month](https://github.com/mozilla/bigquery-etl/blob/eab80c34a99d7a1c3c11213e82545bc12fe8068c/sql/moz-fx-data-shared-prod/mozilla_vpn_derived/all_subscriptions_v1/query.sql#L316), and [I've asked for clarification in Slack](https://mozilla.slack.com/archives/C018SNLQP6C/p1654211937529129)), so to be entirely accurate this would need to be increased to around a month, but in that case when filtering on `is_cohort_complete` the current month's subscriptions would always be excluded, which wouldn't be great.

So I think this is a reasonable approach, especially given that only a very small percentage of subscriptions currently come from Google Play.

sean-rose (approved):

PR #349, opened 2022-06-03
feat: Remove Attribution Dimension Group from active_users_aggr view

kik-kik:
feat: Removed Attribution Dimension Group from active_users_aggr view


lucia-vargas-a (approved):
Thank you

PR #350, opened 2022-06-03
Attempting to fix is_default_browser counting by converting to string values

kik-kik:
Attempting to fix is_default_browser counting by converting to string values.

Otherwise `null` values from the table get counted as `false` resulting in values not representing the known state. The assumption here is that an absence of value, in this case, does not mean we know someone's browser preferences.

### DIFF
--- a/combined_browser_metrics/views/active_users_aggregates.view.lkml
+++ b/combined_browser_metrics/views/active_users_aggregates.view.lkml
@@ -284,6 +284,26 @@ view: +active_users_aggregates {
     sql:  ${TABLE}.active_hours ;;
   }
 
+  dimension: is_default_browser {
+    type:  string
+    case: {
+      when: {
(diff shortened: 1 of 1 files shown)


combined_browser_metrics/views/active_users_aggregates.view.lkml
```diff
@@ -270,6 +270,26 @@ view: +active_users_aggregates {
     sql:  ${TABLE}.active_hours ;;
   }
 
+  dimension: is_default_browser {
```
lucia-vargas-a, line 273: LGTM. Would be great if @Marlene-M-Hirose @ANich can confirm that changing the type from boolean to string in the view doesn't affect your querying in Looker.

lucia-vargas-a (approved):
Double checking in Looker, this would work as a text dimension allowing to filter out nulls instead of counting them as 'False'. The dimension is also not used for counts.

PR #355, opened 2022-06-03
Added forecast region dimension to support vpn foreasting model updates

wwyc:
Added dimension to support vpn forecasting model work.  It has been confirmed with Varun de Sa and Rebecca Yeh from finance that these are the categories they needed.


mozilla_vpn/views/subscription_events_table.view.lkml
```diff
@@ -14,6 +14,17 @@ view: +subscription_events_table {
     sql: CONCAT(${plan_interval_count},"_",  ${plan_interval});;
   }
 
+  dimension: forecast_region {
+    description: "Indicates region used in financial forecasting.  This is to primarily support finance forecasting work."
+    type: string
+    sql: CASE
+          WHEN ${pricing_plan} LIKE "%-eur-%" THEN "Europe"
+          WHEN ${pricing_plan} LIKE "%-chf-%" THEN "Switzerland"
+          WHEN ${pricing_plan} LIKE "1-month-usd-4.99" THEN "Wave_1 monthly legacy only"
```
sean-rose, line 23: No wildcard pattern is used for this one, so it's equivalent to an equality condition.
```suggestion
          WHEN ${pricing_plan} = "1-month-usd-4.99" THEN "Wave_1 monthly legacy only"
```

sean-rose (approved):

PR #370, opened 2022-06-27
Fix subscripiton events delta counts in vpn saasboard

wwyc:
As part of [#3031](https://github.com/mozilla/bigquery-etl/pull/3031), new events were added to the`mozilla_vpn_derived.subscription_events_v1` table for trials data.

The [vpn saasboard - subscriptions growth](https://mozilla.cloud.looker.com/dashboards/416) dashboard relies on subscription events to generate views.
This PR updates the delta counts to consider those events created for trials data.

### DIFF
--- a/mozilla_vpn/views/subscription_events_table.view.lkml
+++ b/mozilla_vpn/views/subscription_events_table.view.lkml
@@ -20,6 +20,6 @@ view: +subscription_events_table {
 
   measure: delta {
     type: sum
-    sql: IF(${event_type} = "New", ${count}, -${count});;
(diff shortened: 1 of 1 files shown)


sean-rose (approved):
This will correct the sign of `delta` for new trial events, but you'll want to be careful how this is used.
- If you're just summing everything by the dimensions listed you're going to double count trials that convert to paid, because those will have both "New Trial" and "New" events.  Maybe having trial vs paid as a dimension could make sense?
- Currently we can't determine the plan interval for Apple trials until/unless they convert to paid, so `plan_interval_type` will end up null for those.

PR #371, opened 2022-06-30
Deng 163 improve query speed active_users_aggregates

lucia-vargas-a:
Adding aggregate awareness to the Explore for active_users_aggregates is part of the improvements to query performance when querying from Looker.

This PR adds two aggregates:
- Current year data for dimensions used in existing Looks under Browsers / <browser_name> / Usage
- Current and last year data for common dimensions identified during the creation of active_users_aggregates


lucia-vargas-a:
Implemented in https://github.com/mozilla/looker-spoke-default/pull/372

PR #372, opened 2022-06-30
Improve query speed active_users_aggregates

lucia-vargas-a:
Adding aggregate awareness to the Explore for active_users_aggregates is one of the improvements to query performance when querying from Looker.

This PR adds two aggregates:

    Current year data for dimensions used in existing Looks under Browsers / <browser_name> / Usage
    Current and last year data for common dimensions identified during the creation of active_users_aggregates

### DIFF
--- a/combined_browser_metrics/combined_browser_metrics.model.lkml
+++ b/combined_browser_metrics/combined_browser_metrics.model.lkml
@@ -20,7 +20,7 @@ explore: active_users_aggregates {
       dimensions: [active_users_aggregates.app_name, active_users_aggregates.submission_date]
       measures: [daily_active_users, weekly_active_users, monthly_active_users, new_profile, ad_click, organic_search_counts, search_counts, search_with_ad, uri_counts, active_hour]
       filters: [
-        active_users_aggregates.submission_date: "after 2022/01/01"
+        active_users_aggregates.submission_date: "this year"
       ]
     }
     materialization: {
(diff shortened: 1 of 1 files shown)


combined_browser_metrics/combined_browser_metrics.model.lkml
```diff
@@ -9,23 +9,54 @@ explore: active_users_aggregates {
...
-  relationship: one_to_one
-  sql_on: ${active_users_aggregates.country} = ${countries.code} ;;
-}
+  join: countries {
+    type: left_outer
+    relationship: one_to_one
+    sql_on: ${active_users_aggregates.country} = ${countries.code} ;;
+  }
+
+  aggregate_table: rollup__active_users_aggregates_2022_usage {
```
lucia-vargas-a, line 18: For fields used in Looks created under Browsers / <browser_name> / Usage

combined_browser_metrics/combined_browser_metrics.model.lkml
```diff
@@ -9,23 +9,54 @@ explore: active_users_aggregates {
...
+      ]
+    }
+    materialization: {
+      sql_trigger_value: SELECT CURRENT_DATE() ;;
+      increment_key: active_users_aggregates.submission_date
+      increment_offset: 1
+    }
+  }
+
+  aggregate_table: rollup__active_users_aggregates_2021_common {
```
lucia-vargas-a, line 33: For fields commonly used in queries, data since 2021.

ANich (approved):
Approved, but I'm not sure why we'd want to use a combination of aggregation techniques here. If these are the commonly used fields, shouldn't we just aggregate them from the upstream/base table?

lucia-vargas-a:
> Approved, but I'm not sure why we'd want to use a combination of aggregation techniques here. If these are the commonly used fields, shouldn't we just aggregate them from the upstream/base table?

I guess this is part of our discussion and proposal of how and when to use each type of aggregate.

My thoughts:
Aggregating in Looker from `unified_metrics` is not feasible as the calculations and logic implemented in `active_user_aggregates` cannot all be replicated using aggregate awareness.
If there is any part that could be replicated, it would imply moving the logic and the data model into Looker, creating a bigger dependency and making it too difficult to query aggregated data from outside Looker.

The purpose of using aggregate awareness here is to improve the performance by generating a subset from `active_user_aggregates`, while keeping the calculations and logic in BigQuery. One of the positive feedback received about this aggregate table is that the structure in Looker reflects the structure in BigQuery, which makes it very clear for all users.

PR #390, opened 2022-08-01
Add pop analysis for desktop and mobile search counts

alekhyamoz:
Resolves : https://mozilla-hub.atlassian.net/browse/RS-340

### DIFF
--- a/search/search.model.lkml
+++ b/search/search.model.lkml
@@ -119,6 +119,5 @@ explore: +desktop_search_counts {
       increment_key: search_clients_engines_sources_daily.submission_date
       increment_offset: 1
     }
-  }
-
+}
 }
--- a/search/search.model.lkml
+++ b/search/search.model.lkml
@@ -83,7 +83,6 @@ explore: business_development_core_search_users_monthly {
 }
 
 explore: +desktop_search_counts {
-
   join: countries {
     type: left_outer
     relationship: one_to_one
@@ -119,5 +118,6 @@ explore: +desktop_search_counts {
       increment_key: search_clients_engines_sources_daily.submission_date
       increment_offset: 1
     }
-}
+  }
+
 }
--- a/search/views/mobile_search_clients_engines_sources_daily.view.lkml
+++ b/search/views/mobile_search_clients_engines_sources_daily.view.lkml
@@ -1,135 +1,198 @@
 include: "//looker-hub/search/views/mobile_search_clients_engines_sources_daily.view.lkml"
 
 view: +mobile_search_clients_engines_sources_daily {
-  dimension_group: submission {
-    sql: ${TABLE}.submission_date ;;
-    type: time
-    timeframes: [
-      raw,
-      date,
-      day_of_year,
-      day_of_month,
-      week,
-      week_of_year,
-      month,
-      month_name,
-      month_num,
-      quarter,
-      year,
-    ]
-    convert_tz: no
-    datatype: date
-  }
+    dimension_group: submission {
+      sql: ${TABLE}.submission_date ;;
+      type: time
+      timeframes: [
+        date,
+        day_of_week_index,
+        day_of_week,
+        day_of_year,
+        day_of_month,
+        week,
+        week_of_year,
+        month,
+        month_name,
+        month_num,
+        quarter,
+        year,
+      ]
+      convert_tz: no
+      datatype: date
+    }
 
-  parameter: choose_breakdown {
-    label: "Choose Grouping (Rows)"
-    view_label: "Date/Period Selection"
-    type: unquoted
-    default_value: "Month"
-    allowed_value: {label: "Month Number" value:"Month"}
-    allowed_value: {label: "Month and Day" value: "Month_Day"}
-    allowed_value: {label: "Week of Year" value: "WOY"}
-    allowed_value: {label: "Day of Year" value: "DOY"}
-    allowed_value: {label: "Day of Month" value: "DOM"}
-    allowed_value: {label: "Day of Week" value: "DOW"}
-    allowed_value: {value: "Date"}
-  }
-  parameter: choose_comparison {
-    label: "Choose Comparison (Pivot)"
-    view_label: "Date/Period Selection"
-    type: unquoted
-    default_value: "Year"
-    allowed_value: {value: "Year" }
-    allowed_value: {value: "Month"}
-    allowed_value: {value: "Week"}
-  }
-  dimension: day_month_abbreviation {
-    type:  date
-    hidden: yes
-    view_label: "Date/Period Selection"
-    convert_tz: no
-    datatype:  date
-    sql: FORMAT_DATE("%b %d", ${TABLE}.submission_date);;
-  }
-  dimension: day_month_number {
-    type:  date
-    hidden: yes
-    view_label: "Date/Period Selection"
-    datatype:  date
-    sql: FORMAT_DATE("%m-%d", ${TABLE}.submission_date);;
-  }
-  dimension: period_over_period_row  {
-    view_label: "Date/Period Selection"
-    label_from_parameter: choose_breakdown
-    type: string
-    order_by_field: sort_by1
-    sql:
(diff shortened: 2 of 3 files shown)


skahmann3 (changes requested):
Will there be documentation somewhere to explain how to use this? I think adding descriptions to each of the fields should suffice (so brief explanations will pop-up if you click on the info button next to each new field under the new PoP menu within the Explore). Relatedly, will you rename the PoP section from "Date/ Period Selection" to something like "Period over Period Analysis Parameters"? I think stakeholders will get confused by the current naming.

Aside from that, this is massively helpful for PoP analyses! Thank you! Going through the desktop explore, I don't think the "Day of Week" option for "Choose Grouping (Rows)" is working: https://mozilla.cloud.looker.com/x/yu6UqiAvJu64P0DE06Lxf0. But the year x month combo worked for me. But before you start working on that ... The available defaults for the PoP row/ pivot are slightly different than those available in the prior Data Studio dashboards. I would expect week can be removed and quarter will need to be added. I'll start a thread in the ROAR channel on desired functionality for stakeholders so we can verify which options should be double-checked before you start implementing.

Marlene-M-Hirose:
> Will there be documentation somewhere to explain how to use this? I think adding descriptions to each of the fields should suffice (so brief explanations will pop-up if you click on the info button next to each new field under the new PoP menu within the Explore). Relatedly, will you rename the PoP section from "Date/ Period Selection" to something like "Period over Period Analysis Parameters"? I think stakeholders will get confused by the current naming.
> 

We will have a video on how to use this code. Now, originally, I called the PoP section "PoP", but George thought it was better to use Date/Period Selection as a name instead. Open to changing the label to what you suggest, @skahmann3

skahmann3:
Thanks, @Marlene-M-Hirose! My concern with the "Date/ Period Selection" label is this will show up for all stakeholders accessing this explore, regardless of whether they are looking for PoP analyses. I would expect some would start trying to use the new filtering section because they think they need to "select dates or periods" not realizing this is for PoP only. We could also go with "Period over Period Analysis: Date/ Period Selection" or some variant that merges the two if preferred

skahmann3:
Also, for ease of reference here is the updated information from stakeholders on the required functionality: 

Stakeholder 1:
1. Year over year for the currently selected period. Example: if current period is August 1 to August 14, 2022 then year over year compares against Aug 1 to Aug 14 2021.
2. Previous period for the currently selected period.
Example 1: if current period is August 1 to August 14 2022 then previous period is July 18-31 2022.
Example 2: if current period is August 1 to August 7 than previous period is July 25-31.

Stakeholder 2 (should be addressed by changes required for 1):
He typically looks at 2-week window, compares it to previous  2-week window, and the same 2-week in previous year.

Marlene-M-Hirose:
Ok, thank you for adding more details, @skahmann3. I'll discuss this with Alekhya and see what we can do!

alekhyamoz:
@skahmann3 @Marlene-M-Hirose  I have modified the PoP implementation to suit the business requirements. 
The  input filters are 
1. `current date range` is the currently selected period
2. Period: period against which we want to compare (could be: Period, week, year, etc)

Also, pre-computed certain measures that might be of use for business. 

Word of caution: The submission date range should cover the entire date range and is different from the current date range. Eg for YoY comparison. The submission date could be `in range` that covers last year's data as well

skahmann3 (changes requested):
I tried to recreate exactly what's in the [Data Studio dashboard](https://datastudio.google.com/u/0/reporting/d62eb11a-3cd1-4879-a096-2fa648721092/page/KvejB), and I'm not getting the expected results in Looker when comparing against [prior year](https://mozilla.cloud.looker.com/x/ZM3vDZqWFFx84f2yiq7NXQ) and [prior period](https://mozilla.cloud.looker.com/x/57JNbfiXfQy4wNSizB8ZVA). The search counts for the current vs previous year seem fine, but the percent change isn't working. When comparing against the prior period, the numbers seem off and the percent change isn't working. Up to you whether we keep percent change in. It's nice to have, but not a must have if the fix is overly complicated

Also, will this explore pull from temporary tables to speed it up? Otherwise, looks like a slow and expensive pull (1TB for prior year comparisons) each time someone opens the forthcoming dashboard. Or not sure if there's a way to pull just the periods we need (e.g. this July, last July, and not the 11 months in between)?

alekhyamoz:
> I tried to recreate exactly what's in the [Data Studio dashboard](https://datastudio.google.com/u/0/reporting/d62eb11a-3cd1-4879-a096-2fa648721092/page/KvejB), and I'm not getting the expected results in Looker when comparing against [prior year](https://mozilla.cloud.looker.com/x/ZM3vDZqWFFx84f2yiq7NXQ) and [prior period](https://mozilla.cloud.looker.com/x/57JNbfiXfQy4wNSizB8ZVA). The search counts for the current vs previous year seem fine, but the percent change isn't working. When comparing against the prior period, the numbers seem off and the percent change isn't working. Up to you whether we keep the percent change in. It's nice to have, but not a must-have if the fix is overly complicated
> 
> Also, will this explore pull from temporary tables to speed it up? Otherwise, looks like a slow and expensive pull (1TB for prior year comparisons) each time someone opens the forthcoming dashboard. Or not sure if there's a way to pull just the periods we need (e.g. this July, last July, and not the 11 months in between)?

@skahmann3  As per your comments and our discussion the pre-calculated measures have been removed. 
After digging around about the discrepancy between Looker and Data Studio, may be related to 
https://docs.telemetry.mozilla.org/datasets/search/search_aggregates/reference.html
Given that we are comparing YoY the difference would be the accumulated difference which is what we might be noticing.

skahmann3:
@alekhyamoz I don't think you pushed the changes noted to the measures. Also, the scale of the discrepancy is likely too large to just be a gotcha. As compared to the [Data Studio](https://datastudio.google.com/u/0/reporting/d62eb11a-3cd1-4879-a096-2fa648721092/page/KvejB), the [Looker Explore](https://mozilla.cloud.looker.com/x/6B7RS8DcMHpfqXW05wz9ub) is suggesting a % change that is about double what we would expect

alekhyamoz:
@skahmann3  Apologies for the inconvenience! I committed the changes now i.e. removed the pre-calculated PoP measure. 
Also, validated the number for both YoY and PoP measures and % changes. The numbers look good albeit not an exact match. Here is the link to the document that includes the numbers for SAP searches specifically (both PoP and YoY): https://docs.google.com/spreadsheets/d/1-h01sQ7Re3xLumDkdbUpEJupsMpzHO1bs-6At4n57v8/edit#gid=449555314

Coming to limiting the submission date filter, I am looking at the options (such as aggregate tables) on how can we make the explore more efficient but could not figure out a way out yet given that this pulls data from daily tables. 
However, I would suggest going ahead with merging the PR while we research. But will leave that decision to you.

skahmann3:
@Marlene-M-Hirose -- I know you mentioned adding a Looker video for PoP, so want to call out the innovation Alekhya added to add a second filter so stakeholders can just pull the months they need and not the 11 months in between. This should speed things up and cut down on costs

@alekhyamoz -- Preethi will be taking over this ticket and doing the final check for DS. She's setting up her Github credentials for Mozilla. Once she does so, she'll need to be added as a reviewer above

pissac17 (commented):
The dashboard looks great and is very helpful! However, when I recreate the overall search metric numbers , the dashboard numbers look very different from what I was expecting, especially for the current vs prior period metrics. This is something that @skahmann3  has  pointed out as well. Is there a reason on why these numbers look different?

pissac17 (approved):

skahmann3 (approved):
Let's hope this let's you merge!

search/search.model.lkml
```diff
@@ -83,7 +83,6 @@ explore: business_development_core_search_users_monthly {
 }
 
 explore: +desktop_search_counts {
```
lucia-vargas-a, line 85: Could be a good opportunity to separate the explores and their aggregate awareness LookML into their own separate files in the [Explores subfolder](https://github.com/mozilla/looker-spoke-default/tree/main/search/explores).

PR #393, opened 2022-08-12
Remove sub events

wwyc:
Subscription events code in the subscriptions explore was no longer correct after the events logic was updated to accommodate free trials data ([DVPN-157](https://mozilla-hub.atlassian.net/browse/DVPN-157)).  Relevant codes are removed from the explore and views that depended on the old logic.

There is ongoing work to improve cancellation metrics ([FXA-5657](https://mozilla-hub.atlassian.net/browse/FXA-5657)).  Dimension related to cancellation reason is removed for now until the data has been standardized.


mozilla_vpn/views/subscriptions.view.lkml
```diff
@@ -190,33 +181,6 @@ view: +subscriptions {
...
-              ${end_date} AS date,
-              "Cancelled" AS type,
-              ${cancel_reason} AS granular_type,
-              -1 AS delta
-            )
-          ],
-          []
-        )
-      );;
-  }
```
sean-rose, line 218: If there's a need for having granular events at the individual subscription level we could potentially add that to the ETL.

sean-rose (approved):
r+wc

I don't have the context to know if this is likely to break other stuff in the Looker dashboards, so I trust you've got things covered in that regard.

mozilla_vpn/views/subscriptions.view.lkml
```diff
@@ -148,15 +148,6 @@ view: +subscriptions {
...
   }
 
-  dimension: cancel_reason {
-    type: string
-    sql: CASE
-      WHEN ${provider} LIKE "Apple Store%" THEN "Cancelled by IAP"
-      WHEN ${canceled_for_customer} THEN "Cancelled by Customer"
-      ELSE "Payment Failed"
-      END;;
-  }
```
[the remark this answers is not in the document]
wwyc: This logic is incorrect now as it doesn't capture the correct cases (eg. refunds and cancellations through support).  When the work around improving churn metrics led by subscriptions platform team is done, we can add the correct logic in again.  I think the logic should be removed altogether.

mozilla_vpn/views/subscriptions.view.lkml
wwyc: I am not aware that we have this need yet.  We can discuss when the need arises.  Thanks.

PR #394, opened 2022-08-16
Add a workaround for Apple Revenue, calculate MRR

birdsarah:
Partially fixes [DVPN-184](https://mozilla-hub.atlassian.net/browse/DVPN-184)

### DIFF
--- a/mozilla_vpn/views/active_subscriptions_table.view.lkml
+++ b/mozilla_vpn/views/active_subscriptions_table.view.lkml
@@ -89,26 +89,7 @@ view: +active_subscriptions_table {
         END);;
   }
 
-
-  measure: annual_recurring_revenue {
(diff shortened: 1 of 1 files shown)


birdsarah:
There's an alternate version of this where I just make MRR = ARR / 12 and I update ARR to include the new estimated apple amount.

birdsarah:
But I do want to calculate other kinds of MRR (Contraction, ....) so will end up needing to write out these formulas anyway.

birdsarah:
@wwyc would appreciate a review.

birdsarah:
I have reviewed the calculations manually offline to make sure I'm happy with them and they look good to me. Also if apple revenue is taken out of the question they are 1/12 of ARR as you would expect.

wwyc (approved):

PR #395, opened 2022-08-18
Updated saasboards lookML

wwyc:
Updated saasboards lookML to match changes in user defined dashboards.
Updated aggregate tables to include new views.

### DIFF
--- a/mozilla_vpn/dashboards/vpn_saasboard_active_subs.dashboard.lookml
+++ b/mozilla_vpn/dashboards/vpn_saasboard_active_subs.dashboard.lookml
@@ -143,11 +143,13 @@
     model: mozilla_vpn
     explore: active_subscriptions
     type: single_value
-    fields: [active_subscriptions.active_date, active_subscriptions.is_max_active_date]
-    filters:
-      active_subscriptions.is_max_active_date: 'Yes'
-    sorts: [active_subscriptions.active_date desc]
-    limit: 3
+    fields: [metadata.last_modified_date]
+    fill_fields: [metadata.last_modified_date]
+    sorts: [metadata.last_modified_date desc]
+    limit: 1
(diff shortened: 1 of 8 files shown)


mozilla_vpn/dashboards/vpn_saasboard_active_subs.dashboard.lookml
```diff
@@ -143,11 +143,13 @@
...
-    fields: [active_subscriptions.active_date, active_subscriptions.is_max_active_date]
-    filters:
-      active_subscriptions.is_max_active_date: 'Yes'
-    sorts: [active_subscriptions.active_date desc]
-    limit: 3
+    fields: [metadata.last_modified_date]
+    fill_fields: [metadata.last_modified_date]
+    sorts: [metadata.last_modified_date desc]
+    limit: 1
+    dynamic_fields: [{category: table_calculation, expression: 'add_days(-1, ${metadata.last_modified_date})',
```
sean-rose, line 150: I'm curious why we're saying the data was last updated the day before the last modified date.

sean-rose (approved):
r+wc

mozilla_vpn/dashboards/vpn_saasboard_active_subs.dashboard.lookml
wwyc: Last_modified_date always shows the current day but the actual day the pipeline refreshes is the day before.  I think @relud may be able to share more insights on this?

relud: this is the most recent date we should see in the data, like how airflow's `{{ds}}` is the logical date not the actual run date

relud: we do this because people were coming to us saying the dashboard data was out of date on, for example 2022-01-03 at 00:30 UTC, because the dashboard only had data up to 2022-01-01

PR #398, opened 2022-08-19
Firefox for Families: add campaign forecast data 

skahmann3:
For campaign monitoring purposes, as relates to [RS-409](https://mozilla-hub.atlassian.net/browse/RS-409)


bani (commented):
LGTM, but I wonder if this really should live as an independent explore, or if we should wait until we have the rest of the campaign data and build one that has both the actuals and forecast.

skahmann3:
@bani Sounds good to me! It would be nice to attach the actuals to this. I'll leave this open until then. And if you need the data in the interim it's stored here `mozdata.analysis.marketing_firefox_for_families_forecasted_expectations`
We can rename the View/ Explore when we update it

bani:
@skahmann3 perfect, thanks!

bani (approved):

bani:
@skahmann3 let's merge this explore and just place the results side by side for simplicity.

PR #404, opened 2022-08-23
Add automated KPI forecast data to Looker

skahmann3:
Adding table with detailed view of forecast output and table with confidence interval output. Also moving this out of the model file so we have the appropriate view and explore files.

### DIFF
--- a/kpi/views/Automated_KPI_Forecast_Confidence_Intervals.view.lkml
+++ b/kpi/views/Automated_KPI_Forecast_Confidence_Intervals.view.lkml
@@ -1,4 +1,4 @@
-include: "//looker-hub/kpi/views/automated_kpi_confidence_intervals.view.lkml"
+include: "//looker-hub/kpi/views/*"
 
 view: +automated_kpi_confidence_intervals {
 
@@ -12,7 +12,7 @@ view: +automated_kpi_confidence_intervals {
     }
 
   dimension_group: submission {
-    sql: ${TABLE}.date ;;
+    sql: CAST(${TABLE}.date as DATE) ;;
     type: time
     timeframes: [
       date,
--- a/kpi/views/Automated_KPI_Forecast_Confidence_Intervals.view.lkml
+++ b/kpi/views/Automated_KPI_Forecast_Confidence_Intervals.view.lkml
@@ -6,28 +6,16 @@ view: +automated_kpi_confidence_intervals {
     label: "Last Forecasted Date"
   }
 
-  dimension: date {
-    hidden: yes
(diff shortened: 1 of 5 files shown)


skahmann3:
@irrationalagent -- Will you review this PR to add the automated KPI forecast data to Looker? I'm adding both the detailed table with the individual component trends and the confidence intervals table

kpi/views/Automated_KPI_Forecast_Confidence_Intervals.view.lkml
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+  dimension: value {
+    hidden:  yes
+    sql: ${TABLE}.value ;;
+  }
+
+  measure: Estimated_Value {
+    type: number
+    description: "Forecasted if after forecast date, observed if before."
```
irrationalagent, line 39: which value is this using if its after the forecast date? 50th percentile? Also, maybe re-phrase: "Forecast Date" to "Forecast Run Date" to match the label

kpi/views/Automated_KPI_Forecasts.view.lkml
```diff
@@ -0,0 +1,54 @@
...
+      month,
+      month_name,
+      quarter,
+      year,
+    ]
+    convert_tz: no
+    datatype: date
+  }
+
+  dimension: forecast_date {
```
irrationalagent, line 20: "Forecast Run Date" to make consistent with the CIs?

irrationalagent (approved):
r+ wc...
would it make sense to join these two explores? or is there something im missing where they wouldn't line up nicely

kpi/views/Automated_KPI_Forecasts.view.lkml
skahmann3: Fixed.

kpi/views/Automated_KPI_Forecast_Confidence_Intervals.view.lkml
skahmann3: Fixed the label. The value isn't the 50th percentile, it's the average of the uncertainty samples. (I've also added this to the label)

skahmann3:
Thanks @irrationalagent! Merging is a bit tricky given the dates from the two datasets. `Automated KPI Forecasts` has the daily forecast data for QDAU and DAU and `Automated KPI Confidence Intervals` rolls up the daily data to monthly. I've updated the two explores to not allow users to switch to a date specification besides day (for the raw output) and month (for the rolled-up output) as we don't want Looker rolling this data up on its own

PR #409, opened 2022-08-31
Create crashes explore for stability & release dashboarding

fbertsch:


firefox_desktop/explores/crashes.explore.lkml
```diff
@@ -0,0 +1,39 @@
+include: "//looker-hub/firefox_desktop/views/crashes_daily.view.lkml"
+include: "../views/crash_usage.view.lkml"
+include: "/shared/views/countries.view.lkml"
+
+
+explore: crash_usage {
+  label: "Crashes"
+  description: "Crash counts for Desktop Firefox, derived from the crash ping."
```
ANich, line 8: nit: even though the explore and view have the same name, I always like to add the `from`/`view_name` parameter so it's 100% clear

firefox_desktop/explores/crashes.explore.lkml
```diff
@@ -0,0 +1,39 @@
...
+    sql_on: ${crashes_daily.client_id} = ${crash_usage.client_id}
+            AND ${crashes_daily.submission_date} = ${crash_usage.submission_date} ;;
+    relationship: one_to_one
+    fields: []
+  }
+
+  join: buildhub {
+    sql_on: ${crash_usage.version} = ${buildhub.version} AND ${crash_usage.channel} = ${buildhub.channel} ;;
+    relationship: many_to_one
+    fields: []
```
ANich, line 20: I'm not following these joins, why `fields: []`?

firefox_desktop/views/crash_usage.view.lkml
```diff
@@ -0,0 +1,369 @@
+view: crash_usage {
+  sql_table_name: mozdata.telemetry.clients_daily ;;
+
+  # We use a Looker Filter so we can apply it directly to both crash_usage and crash views.
+  # Date filter predicate pushdown doesn't seem to be applied through the join (i.e.
+  # when we filter on submission_date here, because we join on submission_date with
+  # the crashes view, we would expect the predicate filter to apply there as well.
```
ANich, line 7: `crashes` = `crashes_daily`?

firefox_desktop/views/crash_usage.view.lkml
```diff
@@ -0,0 +1,369 @@
...
+
+  measure: active_hours {
+    type: sum
+    sql: ${active_hours_sum} ;;
+    description: "Total active hours, derived from active ticks (see the probe dictionary for details)."
+  }
+
+  measure: main_crashes {
+    type: sum
+    sql: ${crashes_daily.main_crash_count} ;;
```
ANich, line 141: I think it's weird that you can reference a different view from this one without an explicit include or something 🤔 Why not define and get these from the view in the above Explore join?

firefox_desktop/views/crash_usage.view.lkml
```diff
@@ -0,0 +1,369 @@
+view: crash_usage {
+  sql_table_name: mozdata.telemetry.clients_daily ;;
```
ANich, line 2: I haven't seen the pattern of multiple views pointing to the same table much but I think it's something to think about. No strong feelings about it though

ANich (approved):
Some patterns I haven't seen before here but it seems to work when I check it out in Looker. Can't hurt to experiment so I left some clarifying questions/nits

firefox_desktop/views/crash_usage.view.lkml
fbertsch: Right! Good catch

firefox_desktop/views/crash_usage.view.lkml
fbertsch: Good point. IMO it simplifies use of this explore, and let's us simplify it for the end users without adding a bunch of complicated fields they will never need.

firefox_desktop/views/crash_usage.view.lkml
fbertsch: Good question. Because what we actually want to calculate is `crash_rate`, which takes a value from both the `crashes_daily` table (total crashes) and `clients_daily` (active hours). `crash_rate = crashes / active_hours`. Given that, we have two options:
- Pull the `active_hours` value into the measure in `crashes_daily`
- Pull the `crash_count` value into the measure in `clients_daily`

I opted for the latter, since that let me define a single view file (and then we use the dimensions defined in `clients_daily`, rather than any of those defined in `crashes_daily`).

firefox_desktop/explores/crashes.explore.lkml
fbertsch: We join these views simply to pull in some data for measures or dimensions, but don't actually want to expose any of the fields in them. For example, the buildhub data just gives us the date the build was published, and we use that to calculate a `days_since_published` field when subtracted from `submission_date`.

PR #412, opened 2022-09-06
remove unused pop unified_metrics

Marlene-M-Hirose:


scholtzan (approved):
I assume this is not used anywhere at the moment?

Marlene-M-Hirose:
No, I think I put it in as a test before Lucia made the active_users_aggregates tables

PR #415, opened 2022-09-09
Bug(DENG-128): Removed a line resulting in laways 180 day window being filtered even when smaller window requested

kik-kik:
# Bug(DENG-128): Removed a line resulting in laways 180 day window being filtered even when smaller window requested


kik-kik:
This will actually not resolve it, seems like this is coming from LookML generator and is already inside `looker-hub`:
https://github.com/mozilla/looker-hub/blob/main/firefox_accounts/explores/event_counts.explore.lkml#L10

kik-kik:
This is correct, closing this PR

PR #421, opened 2022-09-13
added Page Table Explore to MDM Yari

kik-kik:
LookML generation for `page` view run, however, there are some additional fields in `page_table` LookML view file which do not exist inside `page` view which gets exposed an explore. This is manually adding an explore for page_table view.


kik-kik:
Perhaps a better approach would be to manually extend page view in looker-spoke-default and re-define the explore on top of the extended view?

mdn_yari/explores/page_table.explore.lkml
```diff
@@ -0,0 +1,14 @@
+include: "//looker-hub/mdn_yari/views/page_table.view.lkml"
+
+explore: page_table {
```
scholtzan, line 3: Is there anything in this explore that the `page` explore doesn't have?

kik-kik: yes, Florian from MDN was trying to use Looker and noticed the following was missing from the `page` Explore:
https://github.com/mozilla/looker-hub/blob/main/mdn_yari/views/page_table.view.lkml#L335-L351

I assume this is not the first time this happens, I'm not quite sure what is our standard approach to handle this kind of case. In the PR changes, there are two different approaches suggested (creating a new Explore for Page Table or extending the page view to include those dimensions).

Once we agree on how to proceed I will remove the unnecessary changes. I guess one more approach would be to join the two views together in the existing `page` Explore.

Is this something that happens often, and how should cases like this be resolved @scholtzan ?

scholtzan: This is the first time this has happened. So for now this is fine until we come across some other cases where this happens which is when we can think of a more general solution.

scholtzan (approved):

PR #431, opened 2022-10-13
add a derived table/view/explore of fxa aggregates

irrationalagent:
this is meant to be a somewhat temporary solution that can be replaced once the proper refactor of the FxA tables that kik is working on is done. 

for now, it just provides the basic count of FxA flows that start and end in a login or registration. this is what's needed for the MR dashboards.

my intent was to make this as an incremental PDT that updates the past 2 days everyday, in order to catch the flows that are mid- ummm... flow when the job runs.

### DIFF
--- a/firefox_accounts/views/fxa_flow_aggregates.view.lkml
+++ b/firefox_accounts/views/fxa_flow_aggregates.view.lkml
@@ -1,0 +1,150 @@
+# note, this is intended to be a temporary solution for counting registrations and logins in a way that can be segmented by entrypoint and UTM parameters
+# we are working on refactoring the FxA tables more generally.
+
+view: fxa_flow_aggregates {
+  derived_table: {
+    interval_trigger: "24 hours"
+    increment_key: "flow_start_date"
+    # go back 2 days to allow recent flows to complete
+    increment_offset: 2
+    sql: WITH
+          flow_ids AS (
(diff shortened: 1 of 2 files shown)


irrationalagent:
@scholtzan are you OK with this approach? the intention here is that this is a stopgap solution until we make the more permanent changes to the FxA data model that kik is working on. its needed for the MR release monitoring (thus the hard-coding of the date filter)

firefox_accounts/views/fxa_flow_aggregates.view.lkml
```diff
@@ -0,0 +1,174 @@
...
+                  MIN(DATE(timestamp)) AS flow_start_date,
+              FROM firefox_accounts.fxa_content_auth_oauth_events
+              WHERE DATE(timestamp) >= DATE(2022,9,1)
+              AND event_type = 'fxa_email_first - view'
+              GROUP BY 1
+          ),
+
+      flow_agg AS (
+        SELECT
+          f.*,
```
scholtzan, line 32: it would be preferable to explicitly list the columns here and in the GROUP BY, otherwise if a field gets added in the base table this query will break and would need to be fixed manually

scholtzan (approved):
yeah, that seems fine as a temporary solution

PR #432, opened 2022-10-14
Start MR 2022 Folder

danielkberry:

PR #433, opened 2022-10-17
add mr 2022 views

irrationalagent:
This will replace https://github.com/mozilla/looker-spoke-default/pull/432 which i will close shortly.

### DIFF
--- a/firefox_desktop/firefox_desktop.model.lkml
+++ b/firefox_desktop/firefox_desktop.model.lkml
@@ -1,6 +1,7 @@
 connection: "telemetry"
 
 include: "//looker-hub/firefox_desktop/views/*.view.lkml"
+include: "/mr_2022/views/*.view.lkml"
 include: "//looker-hub/firefox_desktop/explores/*.explore.lkml"
 include: "explores/*.explore.lkml"
 include: "views/*.view.lkml"
@@ -19,6 +20,12 @@ explore: +client_counts {
       client_counts.submission_date: "7 days",
       client_counts.sample_id: "0"
     ]
+  }
+  join: felt_privacy_pinned_retention_2_week {
+    type: left_outer
+    relationship: one_to_one
+    sql_on: ${client_counts.client_id} = ${felt_privacy_pinned_retention_2_week.client_id} AND ${client_counts.submission_date} = ${felt_privacy_pinned_retention_2_week.submission_date};;
+
   }
 }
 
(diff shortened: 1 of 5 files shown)


mr_2022/views/felt_privacy_pinned_retention_2_week.view.lkml
```diff
@@ -0,0 +1,73 @@
...
+            PARTITION BY client_ID
+            ORDER BY UNIX_DATE(submission_date)
+            RANGE BETWEEN 7 PRECEDING AND CURRENT ROW
+            ) AS was_pinned_this_week
+        FROM base
+      ), filtered_to_currently_pinned AS (
+        SELECT client_id,
+          submission_date,
+          was_private_pinned,
+          CAST(was_pinned_at_end_of_next_2_weeks AS int) AS was_pinned_at_end_of_next_2_weeks,
```
irrationalagent, line 26: I noticed this was returning NULL for some rows, which is why I coalesce below

firefox_desktop/firefox_desktop.model.lkml
```diff
@@ -20,6 +21,12 @@ explore: +client_counts {
       client_counts.sample_id: "0"
     ]
   }
+  join: felt_privacy_pinned_retention_2_week {
```
irrationalagent, line 24: I decided that since daniel's query runs on clients daily, and returns client_id-submission_date pairs, that we should just join the results back onto the client counts explore since that will allow further segmentation of the users for this measure. Am open to counter-arguments

irrationalagent: @danielkberry [here's an example of how this works in creating the chart](https://mozilla.cloud.looker.com/explore/firefox_desktop/client_counts?qid=WNnjoHYorWJrXJCYUdZPeU&toggle=fil) - you have to check this branch out in looker from the git menu to be able to see it

mr_2022/explores/mr_2022.explore.lkml
```diff
@@ -0,0 +1,20 @@
+include: "../views/*.view.lkml"
+include: "/shared/views/*"
+
+explore: felt_privacy_pinned_retention_2_week {
```
irrationalagent, line 4: I can probably delete this if its not going to be its own explore

mr_2022/views/felt_privacy_pinned_retention_2_week.view.lkml
```diff
@@ -0,0 +1,73 @@
...
+          CAST(was_pinned_this_week AS int) AS was_pinned_this_week
+        FROM with_rolling
+        -- WHERE was_pinned_this_week
+      )
+
+      SELECT submission_date,
```
irrationalagent, line 32: query not longer aggregates

scholtzan (approved):
lgtm

PR #436, opened 2022-10-19
Update crashes explore

fbertsch:
- Use clients_daily_joined_v1
- Create aggregate tables for Stability & Release dashboard

This improves usability in a few key ways.
1. Use clients_daily_joined means we don't have to join in Looker queries
2. Removing the join lets us build aggregate tables where previously we couldn't
3. Aggregate tables improve usability of the dashboard considerably

### DIFF
--- a/firefox_desktop/explores/crashes.explore.lkml
+++ b/firefox_desktop/explores/crashes.explore.lkml
@@ -18,7 +18,7 @@ explore: crash_usage {
     sql_on: ${crash_usage.country_code} = ${countries.code} ;;
   }
 
-  sql_always_where: ${crash_usage.submission_date} >= "2022-09-01" ;;
+  sql_always_where: ${crash_usage.submission_date} >= "2021-01-01" ;;
 
   always_filter: {
     filters: [
--- a/firefox_desktop/explores/crashes.explore.lkml
+++ b/firefox_desktop/explores/crashes.explore.lkml
@@ -1,10 +1,20 @@
 include: "../views/crash_usage.view.lkml"
 include: "/shared/views/countries.view.lkml"
 
+datagroup: clients_daily_joined_partitions {
+  max_cache_age: "24 hours"
+  sql_trigger:
+    SELECT MAX(last_modified_time)
+    FROM `moz-fx-data-shared-prod`.telemetry_derived.INFORMATION_SCHEMA.PARTITIONS
+    WHERE table_name = "clients_daily_joined_v1" ;;
+  label: "Clients Daily Joined Partitions"
+  description: "Updates when new partitions are added to clients_daily_joined"
+}
 
 explore: crash_usage {
   label: "Crashes"
   description: "Crash counts for Desktop Firefox, derived from the crash ping."
+  persist_with: clients_daily_joined_partitions
 
   join: buildhub {
     sql_on: ${crash_usage.version} = ${buildhub.version} AND ${crash_usage.channel} = ${buildhub.channel} ;;
@@ -67,7 +77,7 @@ explore: crash_usage {
       ]
     }
     materialization: {
-      sql_trigger_value: SELECT MAX(submission_date) FROM `moz-fx-data-shared-prod`.telemetry_derived.clients_daily_joined_v1 ;;
+      datagroup_trigger: clients_daily_joined_partitions
       increment_key: "submission_date"
     }
   }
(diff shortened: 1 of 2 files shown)


firefox_desktop/explores/crashes.explore.lkml
```diff
@@ -26,14 +18,240 @@ explore: crash_usage {
...
+      ]
+      filters: [
+        crash_usage.submission_date: "3 month",
+        crash_usage.days_since_build_published: "<=56",
+        crash_usage.ten_percent_sample: "Yes",
+      ]
+    }
+
+    materialization: {
+      sql_trigger_value: SELECT MAX(submission_date) FROM `moz-fx-data-shared-prod`.telemetry_derived.clients_daily_joined_v1 ;;
```
ANich, line 101: This seems pretty expensive for a trigger check (every 5m I think?). Could you use the information schema? Something like [this](https://github.com/mozilla/looker-spoke-default/blob/e1315853507fc1ac6e78d252d53dc8df5f5f322b/mozilla_vpn/explores/active_subscriptions.explore.lkml#L75-L83):

 ```suggestion
       sql_trigger_value: SELECT MAX(last_modified_time) FROM `moz-fx-data-shared-prod`.telemetry_derived.INFORMATION_SCHEMA.PARTITIONS WHERE table_name = "clients_daily_joined_v1";;
```

File: firefox_desktop/explores/crashes.explore.lkml [hunk 2]
```diff
@@ -26,14 +18,240 @@ explore: crash_usage {
...
+        crash_usage.channel: "beta"
+      ]
+    }
+    materialization: {
+      sql_trigger_value: SELECT MAX(submission_date) FROM `moz-fx-data-shared-prod`.telemetry_derived.clients_daily_joined_v1 ;;
+      increment_key: "submission_date"
+    }
+  }
+
+  aggregate_table: full_beta_buildid_rollups {
```
ANich, line 227: This seems almost identical to `full_nightly_buildid_rollups` besides the filter, could that filter just be added as a dimension? Could turn two queries into one.

firefox_desktop/views/crash_usage.view.lkml
```diff
@@ -1,38 +1,27 @@
...
-    type: date
-    default_value: "28 days"
-    description: "Submission date (date that we received the associated crash or usage data from the client)."
-  }
-
-  filter: sample_id {
-    type: number
-    description: "Sample ID to filter on. Use this to speed up iterative analysis."
-  }
+  sql_table_name: `moz-fx-data-shared-prod`.telemetry_derived.clients_daily_joined_v1 ;;
```
ANich, line 2: Isn't `mozdata.telemetry.clients_daily` a downstream view of this table? I haven't looked much at the view definitions but I imagine it changes a lot in between? I don't follow how this removes the join parts

ANich (commented):
LGTM for the most part. Left a few questions including one potentially blocking about the trigger value

firefox_desktop/explores/crashes.explore.lkml
fbertsch: Good call 👍

firefox_desktop/explores/crashes.explore.lkml
fbertsch: It is almost identical. What I'm trying to avoid is a daily rollup of 100% of release, and there's no clear way to do that with a single aggregate table that also lets you specify either nightly _or_ beta. (The problem is that Looker requires filters to exactly match. It's been a real headache trying to make sure these aggregates get used by the tiles in the stability dashboard).

firefox_desktop/explores/crashes.explore.lkml
```diff
@@ -1,18 +1,20 @@
...
   label: "Crashes"
   description: "Crash counts for Desktop Firefox, derived from the crash ping."
-
-  join: crashes_daily {
-    sql_on: ${crashes_daily.client_id} = ${crash_usage.client_id}
-            AND ${crashes_daily.submission_date} = ${crash_usage.submission_date} ;;
-    relationship: one_to_one
-    fields: [main_crash_count, content_crash_count, gpu_crash_count, rdd_crash_count, utility_crash_count, socket_crash_count, vr_crash_count]
-  }
+  persist_with: clients_daily_joined_partitions
```
ANich, line 17: TIL

ANich (approved):
🚢

PR #441, opened 2022-10-26
do not make fxa_flow_aggregates incremental

irrationalagent:
I noticed that my incrementing scheme on this table isn't working, leading to duplication and appending of already existing rows - this leads to inflated numbers when the table is aggregated by the user. so this just rebuilds the whole table every new day now.

I'm not 100% sure what is going on. my understanding, clearly wrong, of what was going was that every 24 hours based on the `flow_start_ts` column, the table would be rebuilt for the previous 2 x 24 hours or 2 days. 

before that, I tried using the flow_start_date column as the `increment_key` with an `increment_offset` of 2. however, that threw an error when the table attempted to update because of an invalid comparison between a TIMESTAMP and DATE. 

my guess is that either the increment offset of `2` is actually being read as "2 HOURS", or there is something else going on.

@ANich can you help me understand the error of my ways here?

Checklist for reviewer:

### DIFF
--- a/firefox_accounts/views/fxa_flow_aggregates.view.lkml
+++ b/firefox_accounts/views/fxa_flow_aggregates.view.lkml
@@ -3,10 +3,7 @@
 
 view: fxa_flow_aggregates {
   derived_table: {
-    interval_trigger: "24 hours"
-    increment_key: "flow_start_ts"
-    # go back 2 days to allow recent flows to complete
-    increment_offset: 2
+    sql_trigger_value: SELECT CURRENT_DATE() ;;
     sql: WITH
           flow_ids AS (
               SELECT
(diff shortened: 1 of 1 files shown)


irrationalagent:
BTW you can see what's going on with the PDT at `mozdata.tmp.LR_4MHTE1666567133670_fxa_flow_aggregates`

irrationalagent:
@scholtzan sorry was trying to avoid tagging you but I just realized @ANich is out

firefox_accounts/views/fxa_flow_aggregates.view.lkml
```diff
@@ -3,10 +3,7 @@
 
 view: fxa_flow_aggregates {
   derived_table: {
-    interval_trigger: "24 hours"
-    increment_key: "flow_start_ts"
```
scholtzan, line 7: I think the problem might be that the SQL query does not have an `incrementcondition`: https://cloud.google.com/looker/docs/reference/param-view-increment-key#creating_an_incremental_sql-based_pdt I basically return the same set of data (plus some data that got added) during every interval.

irrationalagent:
commit message should have been "add increment *condition*"

scholtzan (approved):
I _think_ this should be right

PR #452, opened 2022-11-10
add total ad click and total search with ads measures

mmccorks:
Checklist for reviewer:

### DIFF
--- a/firefox_desktop/views/newtab_interactions.view.lkml
+++ b/firefox_desktop/views/newtab_interactions.view.lkml
@@ -195,6 +195,20 @@ view: newtab_interactions {
     sql: ${tagged_follow_on_search_ad_clicks} ;;
   }
 
+  measure: sum_all_search_ad_clicks {
+    group_label: "Search"
+    label: "Total Ad Clicks"
+    type: sum
+    sql: ${tagged_search_ad_clicks} + ${follow_on_search_ad_clicks} ;;
+  }
+
+  measure: sum_all_search_ad_impressions {
+    group_label: "Search"
+    label: "Total Ad Impressions"
+    type: sum
+    sql: ${tagged_search_ad_impressions} + ${follow_on_search_ad_impressions} ;;
+  }
+
   measure: visits_with_search {
     group_label: "Search"
     type: count_distinct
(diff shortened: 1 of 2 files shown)


firefox_desktop/views/newtab_interactions.view.lkml
```diff
@@ -195,6 +195,20 @@ view: newtab_interactions {
     sql: ${tagged_follow_on_search_ad_clicks} ;;
   }
 
+  measure: sum_all_search_ad_clicks {
+    group_label: "Search"
+    label: "Total Ad Clicks"
+    type: sum
+    sql: ${tagged_search_ad_clicks} + ${follow_on_search_ad_clicks} ;;
```
ANich, line 202: I don't think these are mutually exclusive - see above `sum_tagged_follow_on_search_ad_clicks`, it's defined AS:

```
  COUNTIF(
    is_tagged_search_ad_click
    AND is_follow_on_search_ad_click
  ) AS tagged_follow_on_search_ad_clicks,
```
Same for ad impressions

mmccorks: Ah ok, so this means that follow on is a subset of tagged searches then. This is different than our usual way we define these things, let me make the appropriate adjustments.

mmccorks: I've made some edits, can you take a look @ANich? 
Do we need to update the dimensions definitions in the view as well? And should we add labels to all of these metrics?

ANich:
We should change this upstream, likely here: 

https://github.com/mozilla/bigquery-etl/blob/0647d3ed7a70148eec38a4cfd993914105f5f835/sql/moz-fx-data-shared-prod/telemetry_derived/newtab_interactions_v1/query.sql#L160-L161 

So if I understand right, tagged searches is for searches that are tagged but NOT follow-on?

firefox_desktop/views/newtab_interactions.view.lkml
ANich: See https://github.com/mozilla/looker-spoke-default/pull/452#issuecomment-1325227620, this should probably be updated at the source so it doesn't cause confusion elsewhere. Yeah this could definitely use some descriptions

mmccorks:
Yes, typically in other tables we have `tagged_sap` and `tagged_follow_on` as separate fields. We also generally call ad impressions `search_with_ads` but I think calling it ad impression here is ok to stay consistent with the tile impression data.

mmccorks:
Ok so - I've changed the search definitions so that tagged searches are just tagged searches not including follow-ons to remain consistent with other search tables, added total tagged search which contains both tagged and tagged follow on.
I removed untagged follow on searches with ads/untagged follow on ad clicks as these should not be any different than tagged follow on (we do not have follow on searches with ads or ad clicks instrumented for non-partner search engines) and to reduce confusion.

ANich (approved):

PR #462, opened 2022-12-02
edit sql to avoid overcouting clients on event-based measures

florezita:
Note: editing sql to only provide measures across all colors in the colorways collection (not by individual color). Otherwise, in the event-based measures, we are over-counting user engagement. For example, in the case that one user "explores" or "plays" with multiple colorways in the modal, in the aggregate the one user should count as 1. Previously, the user would count once if it interacted with one colorway and a second time if it interacted with a second colorway.

Checklist for reviewer:

### DIFF
--- a/firefox_desktop/views/colorways.view.lkml
+++ b/firefox_desktop/views/colorways.view.lkml
@@ -24,7 +24,6 @@ WITH indp_voices_colorways AS
          "activist-soft-colorway@mozilla.org"])),
  currently_set AS
   (SELECT DATE(submission_timestamp) AS submission_date,
-          SPLIT(environment.addons.theme.id, '-colorway')[OFFSET(0)] AS id,
           100 * COUNT(DISTINCT client_id) AS clients_currently_set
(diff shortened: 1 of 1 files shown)


florezita:
Yea-- i don't mind losing the functionality of splitting it by colorway. They are after total DAU and i doubt anyone has made charts.

The only thing extra I need to think about is capturing all the ways that people can interact with exploring or choosing colorways. I only have the entry point being modal here and to have a more complete understading, I do think I should add about:addons and onboarding numbers.

Should I close this pull request, edit and create a new PR for when it is more complete then?

irrationalagent:
I think it makes sense to do it all in one PR but you can continue to use this one if you want

irrationalagent:
I think for your purposes you could use either. Both will fire when a colorway is set, though the `coloways_modal` events contain more detail.

The main reason to use the `coloways_modal` is when you need to distinguish between a colorways being set for the first time (method = `set_colorway`) or changed (method = `change_colorway`). Or, if you need to know whether the colorway was set from about:addons or firefox view (the event `object` will tell you that)

BTW tip: if you want to test when events are generated, you can (e.g.) set the colorway on your own browser then navigate to about:telemetry. then click on the events tab on the left.

irrationalagent (approved):
looks good to me.

PR #464, opened 2022-12-10
DENG 580 - updated dimensions to use GB format

wwyc:
This lookml is used for the Data Platform Usage Monitoring dashboards.  It is currently being actively developed and tested.  The current phase is poc and will likely require more iterations of changes.  The impact and risk is low for reviewing this PR.

[DENG 580](https://mozilla-hub.atlassian.net/browse/DENG-580):  Updated fields related to storage size in unit of GB.  This is because storage costs are in the unit of GB.

* connected table storage explores with table inventory to allow filtering by table type

* destination tables with partitions were showing up in job usage table as separate rows.  Destination table name and partition were extracted to separate dimensions to allow grouping by destination tables names without partition.

Checklist for reviewer:

### DIFF
--- a/monitoring/views/bigquery_usage.view.lkml
+++ b/monitoring/views/bigquery_usage.view.lkml
@@ -18,6 +18,16 @@ view: +bigquery_usage {
     type: string
   }
 
+  dimension: destination_table_id {
+    sql: split(destination_table_id, '$')[SAFE_OFFSET(0)];;
+    type: string
+  }
+
+  dimension: partition {
+    sql: split(destination_table_id, '$')[SAFE_OFFSET(1)];;
+    type: string
+  }
+
   measure: avg_total_job_cost_usd{
     type: average_distinct
     sql_distinct_key: ${job_id} ;;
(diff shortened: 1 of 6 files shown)


monitoring/views/bigquery_usage.view.lkml
```diff
@@ -18,6 +18,16 @@ view: +bigquery_usage {
     type: string
   }
 
+  dimension: destination_table_id {
+    sql: split(destination_table_id, '$')[SAFE_OFFSET(0)];;
+    type: string
+  }
+
+  dimension: partition {
```
kik-kik, line 26: Should we maybe also make this a date field? Could process the extracted value with `PARSE_DATE` and that would allow us to do `partition` field date-based filtering. Here's an example of what I mean:

https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/firefox_accounts_derived/fxa_auth_events_v1/query.sql#L32

kik-kik (approved):
Looks good, left just one comment. Please read it before merging.

monitoring/views/bigquery_usage.view.lkml
wwyc: It looks like not all partitions are the same format.  I will leave this field as string format until there is a need to adjust it.

PR #486, opened 2023-01-30
DO-947, DENG-784 Rollups and add topsites metrics for Newtab interactions 

ANich:
RE: https://mozilla-hub.atlassian.net/browse/DO-947, https://mozilla-hub.atlassian.net/browse/DENG-784

Introducing some pre-aggregations for newtab Looker performance. I did a best effort at picking dimensions and measures that are used in dashboards, tagging folks for a second opinion.

----

Checklist for reviewer:

### DIFF
--- a/firefox_desktop/explores/newtab_interactions.explore.lkml
+++ b/firefox_desktop/explores/newtab_interactions.explore.lkml
@@ -1,8 +1,9 @@
 include: "../views/newtab_interactions.view.lkml"
 include: "../../shared/views/countries.view.lkml"
-include: "//looker-hub/firefox_desktop/datagroups/newtab_interactions_v1_last_updated.datagroup.lkml"
+include: "//looker-hub/firefox_desktop/datagroups/newtab_interactions_v1_last_updated.datagroup"
 
 explore: newtab_interactions {
+  persist_with: newtab_interactions_v1_last_updated
   sql_always_where: ${newtab_interactions.submission_date} >= '2022-07-01' ;;
   label: "New Tab Interactions"
   from: newtab_interactions
--- a/firefox_desktop/explores/newtab_interactions.explore.lkml
+++ b/firefox_desktop/explores/newtab_interactions.explore.lkml
@@ -1,9 +1,8 @@
 include: "../views/newtab_interactions.view.lkml"
 include: "../../shared/views/countries.view.lkml"
-include: "//looker-hub/firefox_desktop/datagroups/newtab_interactions_v1_last_updated.datagroup"
+include: "//looker-hub/firefox_desktop/datagroups/newtab_interactions_v1_last_updated.datagroup.lkml"
 
 explore: newtab_interactions {
-  persist_with: newtab_interactions_v1_last_updated
   sql_always_where: ${newtab_interactions.submission_date} >= '2022-07-01' ;;
   label: "New Tab Interactions"
   from: newtab_interactions
@@ -32,6 +31,8 @@ explore: newtab_interactions {
         newtab_interactions.pocket_sponsored_stories_enabled,
         newtab_interactions.topsites_enabled,
         countries.tier,
+        countries.name,
+        countries.pocket_available_on_newtab,
       ]
       measures: [
         newtab_interactions.visits,
@@ -41,13 +42,7 @@ explore: newtab_interactions {
     }
 
     materialization: {
-      sql_trigger_value:
-      SELECT
-        MAX(last_modified_time)
-      FROM
-        moz-fx-data-shared-prod.telemetry_derived.INFORMATION_SCHEMA.PARTITIONS
-      WHERE
-        table_name = "newtab_interactions_v1";;
+      datagroup_trigger: newtab_interactions_v1_last_updated
       increment_key: newtab_interactions.submission_date
       increment_offset: 1
     }
@@ -61,6 +56,8 @@ explore: newtab_interactions {
         newtab_interactions.country_code,
         newtab_interactions.pocket_story_position,
         countries.tier,
+        countries.name,
+        countries.pocket_available_on_newtab,
       ]
       measures: [
         newtab_interactions.visits_with_pocket_impressions,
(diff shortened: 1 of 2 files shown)


irrationalagent:
I'll defer to @mmccorks on the dimensions / measures question. This seems to cover the ones i used in my most recent dashboard.

Also, for validation's sake, when we run a query through the explore once this change merges, is there an easy way to see from the UI that the rollup (rather than the source table) was used? Does it just show up in the SQL tab like `SELECT blah FROM rollup__newtab_interactions_topsites_per_day`?

Finally, out of curiosity, can the aggregate awareness stuff work with tables that are defined separately from looker, e.g. in BQ ETL? I.e. could I in theory tell looker, "this table that already exists is a rollup of this explore, use that when you can?" (I am NOT saying that is what we should do here, I'm just trying to understand what's possible).

ANich:
> Does it just show up in the SQL tab like `SELECT blah FROM rollup__newtab_interactions_topsites_per_day`?

Yup, it shows up in the SQL tab and also at the top-right where the amount of data scanned shows up. (Scanned x instead of y using aggregate table)


> Finally, out of curiosity, can the aggregate awareness stuff work with tables that are defined separately from looker, e.g. in BQ ETL? I.e. could I in theory tell looker, "this table that already exists is a rollup of this explore, use that when you can?" (I am NOT saying that is what we should do here, I'm just trying to understand what's possible).

Not that I'm aware of, though it has been a [feature request](https://portal.feedback.us.pendo.io/app/#/case/64941?currentProductId=53ee6715-9dcb-471f-bd41-7846ac4d9e41&section=requests). The closest we could get to this is using something like materialized views since bigquery can automatically rewrite your query when it matches but there's some challenges there as well

firefox_desktop/explores/newtab_interactions.explore.lkml
```diff
@@ -17,4 +17,182 @@ explore: newtab_interactions {
...
+        newtab_interactions.sum_topsite_clicks,
+        newtab_interactions.sum_topsite_impressions,
+        newtab_interactions.sum_sponsored_topsite_clicks,
+        newtab_interactions.sum_sponsored_topsite_impressions,
+        newtab_interactions.clients_with_topsite_click,
+        newtab_interactions.clients_with_topsite_impression,
+        newtab_interactions.clients_with_sponsored_topsite_click,
+        newtab_interactions.clients_with_sponsored_topsite_impression,
+      ]
+      filters: [newtab_interactions.submission_date: "after 2022-07-01"]
```
mmccorks, line 182: Can we also filter this to only countries with Topsites enabled?

firefox_desktop/explores/newtab_interactions.explore.lkml
```diff
@@ -17,4 +17,182 @@ explore: newtab_interactions {
...
+        newtab_interactions.clients_with_pocket_saves,
+        newtab_interactions.clients_with_pocket_clicks,
+        newtab_interactions.clients_with_organic_pocket_impressions,
+        newtab_interactions.clients_with_organic_pocket_saves,
+        newtab_interactions.clients_with_organic_pocket_clicks,
+        newtab_interactions.clients_with_sponsored_pocket_impressions,
+        newtab_interactions.clients_with_sponsored_pocket_saves,
+        newtab_interactions.clients_with_sponsored_pocket_clicks,
+      ]
+      filters: [newtab_interactions.submission_date: "after 2022-07-01"]
```
mmccorks, line 94: Can we also also add a filter here to only include countries with Pocket?

mmccorks (approved):
Looks good to me, I suggested adding filters for the topsites and pocket countries but it's not a blocker.

ANich:
Pausing this until https://github.com/mozilla/bigquery-etl/pull/3581 and related issues are resolved

ANich:
I paused this anticipating a few column additions so we could do (only) one pre-aggregation run. I'll re-configure this for the columns added in https://github.com/mozilla/bigquery-etl/pull/3702 and merge. When columns need to be added at a later point we' can modify this

firefox_desktop/explores/newtab_interactions.explore.lkml
ANich: We don't currently have a list of those iirc

firefox_desktop/explores/newtab_interactions.explore.lkml
irrationalagent: +1 if its possible

firefox_desktop/explores/newtab_interactions.explore.lkml
```diff
@@ -19,4 +19,174 @@ explore: newtab_interactions {
...
+        newtab_interactions.clients_with_tagged_follow_on_search_ad_impression,
+        newtab_interactions.clients_with_tagged_follow_on_search_ad_click,
+        newtab_interactions.clients_with_any_ad_click,
+        newtab_interactions.clients_with_any_ad_impression,
+      ]
+      filters: [newtab_interactions.submission_date: "after 2022-07-01"]
+    }
+
+    materialization: {
+      datagroup_trigger: newtab_interactions_v1_last_updated
```
irrationalagent, line 145: just for the benefit of my understanding, where does this live, or how is it generated etc?

ANich: These are generated in looker-hub by looking at a table's last updated timestamp (usually by ETL). Here's the most relevant doc: https://docs.telemetry.mozilla.org/cookbooks/looker/performance_caching.html, though in the caching context, and the [proposal](https://docs.google.com/document/d/1dPwa5KZnKHzYj3hULHy9unczCXugwr6rZHW1gveQ2fc/edit#heading=h.5x0d5h95i329).

firefox_desktop/explores/newtab_interactions.explore.lkml
```diff
@@ -19,4 +19,174 @@ explore: newtab_interactions {
...
+        newtab_interactions.pocket_sponsored_stories_enabled,
+        newtab_interactions.topsites_enabled,
+        newtab_interactions.topsites_rows,
+        countries.tier,
+        countries.name,
+        countries.pocket_available_on_newtab,
+      ]
+      measures: [
+        newtab_interactions.visits,
+        newtab_interactions.clients,
```
ANich, line 41: I'll keep an eye on this. The documentation [here](https://cloud.google.com/looker/docs/aggregate_awareness#approximate_distinct) mentions that distinct counts are only pre-aggregated if all the dimensions are included but an exception is approximate distincts. These are all approximate distincts but the documentation is unclear on whether this will be pre-aggregated in an example say only using one or two of these dimensions.

ANich: This doesn't seem to be working, will follow up in https://mozilla-hub.atlassian.net/browse/DS-2592

PR #508, opened 2023-02-24
feat(DENG-713): tweaked fxa_flow_aggregates looker view to use fxa_all_events instead of fxa_content_auth_* bq view

kik-kik:
# feat(DENG-713): tweaked fxa_flow_aggregates looker view to use fxa_all_events instead of fxa_content_auth_* bq view

This is part of the work to have a single view for accessing fxa logs.

### DIFF
--- a/firefox_accounts/views/fxa_flow_aggregates.view.lkml
+++ b/firefox_accounts/views/fxa_flow_aggregates.view.lkml
@@ -20,9 +20,10 @@ view: fxa_flow_aggregates {
                   ANY_VALUE(ua_version) AS browser_version,
                   ANY_VALUE(os_name) AS os_name,
                   MIN(DATE(timestamp)) AS flow_start_date,
-              FROM firefox_accounts.fxa_content_auth_oauth_events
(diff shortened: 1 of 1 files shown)


akkomar (approved):
LGTM, event_category filter added here seems to be matching [this deprecated view](https://github.com/mozilla/bigquery-etl/blob/ae68eec512acb44a88aed05e425ead39240430d8/sql/moz-fx-data-shared-prod/firefox_accounts/fxa_content_auth_oauth_events/view.sql#L25)

irrationalagent (approved):
thank you this is probably what i should have done from the beginning.

irrationalagent:
the intent is to capture any registration or login events, regardless of log, so if there is another log (besides these three) now that might contain those specific events (I'm not aware of any but my knowledge is very out of date here) then we should include it. is the reason for the filter to speed things up?

kik-kik:
ok, quickly double checked and it appears all the reg and login events come from either `auth` or `content`.

PR #512, opened 2023-03-02
Add link to platform health dashboard  - Bug 1816488 

lelilia:
[Bug 1816488](https://bugzilla.mozilla.org/show_bug.cgi?id=1816488)

Adding a new view in order to be able to create a column with links 

This is a first try. I'm not sure if this is the optimal approach but I can't create a link in a custom field, so this is the best option I found, but I would be happy to learn better approaches. 

Checklist for reviewer:


monitoring/views/all_schema_errors_last_14_days.view.lkml
```diff
@@ -0,0 +1,62 @@
+view: all_schema_errors_last_14_days {
```
scholtzan, line 1: I don't think creating a new view is necessary here. Instead, we should extend the existing view that lives here: https://github.com/mozilla/looker-hub/blob/main/monitoring/views/schema_error_counts.view.lkml

The view already gets imported in `[monitoring/monitoring.model.lkml](https://github.com/mozilla/looker-spoke-default/pull/512/files#diff-cd6cbe20333f36b3dbc346f70a6c6fbe04f91f049d8d785473d25373640bb37e)`. It should be possible to add the `bug_link` dimension by adding the following either to `[monitoring.model.lkml](https://github.com/mozilla/looker-spoke-default/pull/512/files#diff-cd6cbe20333f36b3dbc346f70a6c6fbe04f91f049d8d785473d25373640bb37e)` directly or to a separate view file:
```
view: +schema_error_counts {
  dimension: bug_link {
    label: "Bug"
    type: string
    sql:
      CASE
        WHEN ${bug} IS NOT NULL THEN CONCAT("Bug ", SPLIT(${bug}, "=")[OFFSET(1)])
        ELSE "create Bug"
      END

          ;;
    link: {
      label: "{% if value != \"create Bug\" %} view on BugZilla {% else %} create new Bug {% endif %}"
      url: "
        {% if value != \"create Bug\" %}
          {{ bug }}
        {% else %}
          https://bugzilla.mozilla.org/enter_bug.cgi?product=Data%20Platform%20and%20Tools&component=General&bug_status=NEW&bug_type=defect&short_desc=Investigate+schema+error+in+%60{{ document_namespace }}.{{ document_type }}%60+for+%60{{ path | encode_uri }}%60
        {% endif %}"
      icon_url: "https://bugzilla.mozilla.org/favicon.ico"
    }
}
```

lelilia: The problem I have is that I don't have all the information before joining the `schema_error_counts` with the `schema_error_notes`

If I only add a link to the `schema_error_counts` it would create a `create a Bug` link even for issues that already have a Bug 

And I tried to create a link in the `schema_error_counts` and the `schema_error_notes` and then create a COALESCE(schema_error_notes.bug, schema_error_counts.new_bug) ... but then it looses the link ability and just becomes a string

scholtzan: I see what you mean. I think to reduce the complexity here and not having to create multiple views (which we'd need to do for the other checks as well), it's probably fine to have two columns: one showing the link to the bug and one always showing "Create Bug" (even if a bug already exists). 
This way just adding a dimension for the Bug Link to the view would be enough. And we'd not have to maintain a chain of `schema_error_count` views.

lelilia: Good point. I added links to all the views that are used to create the looks with notes now. Once they are landed I could all those columns to the looks.

monitoring/views/distinct_docids_notes.view.lkml
```diff
@@ -0,0 +1,28 @@
...
+    label: "Bug"
+    type: string
+    sql:
+      CASE
+        WHEN ${bug} IS NOT null THEN CONCAT("Bug ", SPLIT(${bug}, "=")[OFFSET(1)])
+        ELSE ""
+      END
+      ;;
+    link: {
+      label: "{% if value != '' %} view on BugZilla {% endif %}"
```
fbertsch, line 20:
```suggestion
      label: "{% if value != '' %} view on Bugzilla {% endif %}"
```

fbertsch (approved):
LGTM. You may have time to get some of the dashboard updated

PR #533, opened 2023-04-03
[DS-2566] Add explore and view for history sidebar

m-d-bowerman:
Adds a view and an explore based on that view that includes fields related to history and bookmark sidebar interactions.

Checklist for reviewer:

### DIFF
--- a/firefox_desktop/views/history_sidebar_clients_daily.view.lkml
+++ b/firefox_desktop/views/history_sidebar_clients_daily.view.lkml
@@ -11,7 +11,7 @@ view: history_sidebar_clients_daily {
   dimension: submission_date {
     type: date
     sql: CAST(${TABLE}.submission_date AS TIMESTAMP);;
-    description: "Submission date of the crash"
+    description: "Submission date of the interaction"
   }
 
   dimension: major_version {
--- a/firefox_desktop/views/history_sidebar_clients_daily.view.lkml
+++ b/firefox_desktop/views/history_sidebar_clients_daily.view.lkml
@@ -1,6 +1,8 @@
+include: "//looker-hub/firefox_desktop/views/clients_daily_joined_table.view"
+
 view: history_sidebar_clients_daily {
 
-  sql_table_name: `moz-fx-data-shared-prod`.telemetry_derived.clients_daily_joined_v1 ;;
+  extends: [clients_daily_joined_table]
 
   dimension: client_id {
     type: string
@@ -8,10 +10,19 @@ view: history_sidebar_clients_daily {
     hidden: yes
   }
 
-  dimension: submission_date {
-    type: date
-    sql: CAST(${TABLE}.submission_date AS TIMESTAMP);;
-    description: "Submission date of the interaction"
+  dimension_group: submission {
+    sql: ${TABLE}.submission_date ;;
+    type: time
+    timeframes: [
+      raw,
+      date,
+      week,
+      month,
+      quarter,
+      year,
+    ]
+    convert_tz: no
+    datatype: date
   }
 
   dimension: major_version {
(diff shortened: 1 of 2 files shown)


firefox_desktop/views/history_sidebar_clients_daily.view.lkml
```diff
@@ -0,0 +1,86 @@
+view: history_sidebar_clients_daily {
+
+  sql_table_name: `moz-fx-data-shared-prod`.telemetry_derived.clients_daily_joined_v1 ;;
```
irrationalagent, line 3: is there a reason you are referencing the _v1 version of this table? we usually use the views where possible, e.g. `mozdata.telemetry.clients_daily`

Also, we have in the past preferred the extension of things that are already in looker-hub, [and it looks like there is already a view for clients_daily_joined here](https://github.com/mozilla/looker-hub/blob/main/firefox_desktop/views/clients_daily_joined_table.view.lkml).

[Here is an example of how that kind of extension can work](https://github.com/mozilla/looker-spoke-default/blob/main/firefox_desktop/views/clients_daily_search.view.lkml#L1-L4). 

[Here are docs on extending views](https://cloud.google.com/looker/docs/reference/param-view-extends)

I could also be missing something here - is there is a reason why you need to have it the way you do?

irrationalagent (changes requested):
This seems like it could work in theory, but I'd like to understand the reasoning for not extending the existing views as I mention in the other comment. 

Also, did you try this out on a branch in looker development mode to make sure that it works to  build the types of charts you need? I notice you don't have any client count measures here, so you are just looking for e.g. "total number of sidebar opens for the whole population"? rather than per-client numbers? Or maybe you are planning to join it to another explore to do that.

firefox_desktop/views/history_sidebar_clients_daily.view.lkml
```diff
@@ -0,0 +1,86 @@
...
+
+  dimension: client_id {
+    type: string
+    sql: ${TABLE}.client_id ;;
+    hidden: yes
+  }
+
+  dimension: submission_date {
```
irrationalagent, line 11: you can optionally extend this field to allow for timeframes other than dates, if you want.

[here is an example](https://github.com/mozilla/looker-hub/blob/main/firefox_desktop/views/clients_daily_joined_table.view.lkml#L1968-L1981)f

firefox_desktop/views/history_sidebar_clients_daily.view.lkml
m-d-bowerman: I didn't want all of the fields brought in from the `clients_daily_joined_table` view, which is why I was referencing the SQL table. I've changed it to extend the view, though -- having those fields in the future is probably worthwhile anyways.

firefox_desktop/views/history_sidebar_clients_daily.view.lkml
m-d-bowerman: Thanks! Added

firefox_desktop/views/history_sidebar_clients_daily.view.lkml
irrationalagent: OK I see - there are a lot of fields in the source table, and it seems like you don't need too many of them. 

[You can always hide the ones you don't want](https://community.looker.com/lookml-5/hiding-fields-from-an-extended-view-2755?postid=7310#post7310) - all you have to do is set the `hidden` field to `yes` for the fields that you don't want to include from the source view. 

Alternatively, if you feel that makes the lookml too bloated (many lines just to hide each field; I don't think there's another way) then I would be OK with you just referencing the table directly (though the view, not the materialized version like you originally had).

I'll leave it up to you.

irrationalagent (approved):

PR #536, opened 2023-04-12
Rename data-quality whiteboard tag to dataquality

scholtzan:
We have to rename the `[data-quality]` whiteboard tag to `[dataquality]` for jbi to sync issues between bugzilla and jira (it doesn't support dashes in tag names...)


lelilia (approved):
lgtm

PR #544, opened 2023-05-02
Adjusted the dau goal calculations

soGaussian:
Checklist for reviewer:

I remember Anna pulled you into a PR that I assigned to her. For this PR, I simply added a new measure in the view that I added.

### DIFF
--- a/duet/views/kpi_dau.view.lkml
+++ b/duet/views/kpi_dau.view.lkml
@@ -128,11 +128,34 @@ view: kpi_dau {
     filters: [period_filtered_measures: "last"]
   }
 
+  measure: dau_goal {
+    view_label: "KPI filtered metrics"
+    type: sum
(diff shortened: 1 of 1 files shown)


lucia-vargas-a (approved):
The SUM on DAU is correct and the syntax looks good.
Since the targets are likely to be adjusted every certain period e.g. quarterly or yearly, I'd recommend that you consider storing them in BigQuery and making them available in Looker. This way you wouldn't need to update the Looker view or any code each time there are new targets but they would be available immediately. It would also enable querying DAU over different target value, analyze the how they affect the metrics and use them in different views without risking a mismatch.

duet/views/kpi_dau.view.lkml
```diff
@@ -128,13 +128,36 @@ view: kpi_dau {
     filters: [period_filtered_measures: "last"]
   }
 
+  measure: dau_goal {
+    view_label: "KPI filtered metrics"
```
lucia-vargas-a, line 132: Could we make this name more explicit? E.g. KPI_metrics_with_target or KPI_target_metrics?

PR #555, opened 2023-05-25
DENG-784 - Version 1 of newtab_visits 

ANich:
Including topsites and top-level metrics/dimensions, including `had_non_impression_engagement` for counting some visits

This will eventually replace `newtab_interactions` - will add pocket and search metrics as a fast-follow but this is to unblock some immediate needs.


mmccorks (approved):
Looks good to me!

PR #560, opened 2023-06-05
View for activation/$ for UAC android campaigns

bani:
Checklist for reviewer:

### DIFF
--- a/duet/views/google_uac_android_activation.view.lkml
+++ b/duet/views/google_uac_android_activation.view.lkml
@@ -1,0 +1,123 @@
+view: google_uac_android_activation {
+
+  derived_table: {
+    sql: WITH uac AS (
+         SELECT
+           date,
+           REGEXP_REPLACE(campaign_name, '(.*) - NULL', '\\1') as campaign,
+           SUM(spend) as cost
+         FROM
+           mozdata.google_ads.daily_campaign_stats
+         WHERE
+            campaign_name NOT LIKE '%iOS%'
+            AND date >= '2022-12-01'
+         GROUP BY
+            1, 2
+        ), activation AS (
+         SELECT
+           first_seen_date as date,
+           REGEXP_REPLACE(adjust_campaign, '([\\w]+).*', '\\1') as campaign,
+           SUM(activated) as activated,
+           COUNT(*) as new_profiles,
+         FROM `moz-fx-data-shared-prod.fenix.new_profile_activation`
+         WHERE submission_date >= '2022-12-01'
+           AND first_seen_date >= '2022-12-01'
+           AND adjust_network = 'Google Ads ACI'
+           AND adjust_campaign LIKE 'Mozilla_%'
+         GROUP BY 1, 2
+        )
+        SELECT
+          date,
+          campaign,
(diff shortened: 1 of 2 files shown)


duet/views/google_uac_android_activation.view.lkml
```diff
@@ -0,0 +1,123 @@
...
+           AND adjust_campaign LIKE 'Mozilla_%'
+         GROUP BY 1, 2
+        )
+        SELECT
+          date,
+          campaign,
+          CASE WHEN campaign LIKE '%_US_%' OR campaign LIKE '%_CA_%' OR campaign LIKE '%NA%' THEN 'NA' ELSE 'EU' END AS region,
+          REGEXP_EXTRACT(campaign, '.*(Event\\d).*') as event,
+          REGEXP_EXTRACT(campaign, '.*(Group\\d).*') as group_number,
+          REGEXP_EXTRACT(campaign, 'Mozilla_FF_UAC_[\\w]{2}_([\\w]{2})_[\\w]{2}_Group\\d_Event\\d') as country_code,
```
fbertsch, line 35: I understand that this is only available on the newest campaigns, but it might not be a good idea to include it here - there's too many caveats. If, for example, someone filters for `date >= "2023-01-01" AND country_code = "DE"`, they will be filtering _out_ DE data from 2023-01-1 through 2023-05-01 (thereabouts), though of course that DE data would be combined w/ other countries

duet/views/google_uac_android_activation.view.lkml
```diff
@@ -0,0 +1,123 @@
...
+    type: string
+    sql: ${TABLE}.campaign ;;
+  }
+
+  dimension_group: campaign_date {
+    description: "Date of campaign spend"
+    type: time
+    datatype: date
+    timeframes: [date, week, month, year]
+    sql: ${TABLE}.date ;;
```
fbertsch, line 81: We usually add this
```suggestion
    sql: ${TABLE}.date ;;
    convert_tz: no
```

duet/views/google_uac_android_activation.view.lkml
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  measure: new_profiles {
```
fbertsch, line 84: appending `_count` might make this (and `activated` -> `activated_count`) easier to understand (especially b/c it's a `sum` measure, which might be confusing at first glance).

fbertsch (approved):
r+wc

duet/views/google_uac_android_activation.view.lkml
bani: What if I rename to "campaign country code"? I did explicitly include it in the description of the field in the explore, but I understand there is a risk people will miss it.
This field is important because it was something Sally was using in our working session. I think it's less risky than people trying to guess by for example creating a filter with campaign name containing "DE" without realizing it's a language for campaigns in Switzerland for example.

fbertsch: Hm, good point. It's a small user base so we can try to make sure it's being used correctly. `campaign_country_code` seems 👍

PR #568, opened 2023-06-16
VPN-4726 - Hide Glean.js data from Mozilla VPN views

brizental:
Checklist for reviewer:

### DIFF
--- a/mozilla_vpn/explores/events.explore.lkml
+++ b/mozilla_vpn/explores/events.explore.lkml
@@ -1,9 +1,12 @@
 include: "//looker-hub/mozilla_vpn/explores/events.explore.lkml"
 
-explore: +event_counts {
+explore: event_counts {
   sql_always_where:  ${events.submission_date} >= '2010-01-01'
-    AND (
-      ${events.metadata__header__x_telemetry_agent} LIKE "%Kotlin%"
-      OR ${events.metadata__header__x_telemetry_agent} LIKE "%JS%"
-    ) ;;
+      -- For version below 2.16 Kotlin and JS were the main Glean integrations
+      (CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_OFFSET(1)] AS INTEGER) < 2 OR (CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_OFFSET(1)] AS INTEGER) = 2 AND CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_OFFSET(2)] AS INTEGER) < 16 )
+        AND (${events.metadata__header__x_telemetry_agent} LIKE "%Kotlin%" OR ${events.metadata__header__x_telemetry_agent} LIKE "%JS%"))
+      OR
+      -- For version above 2.16 Kotlin, Swift and Rust are the main Glean integrations
+      (CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_OFFSET(1)] AS INTEGER) > 2 OR (CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_OFFSET(1)] AS INTEGER) = 2 AND CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_OFFSET(2)] AS INTEGER) >= 16)
+        AND (${events.metadata__header__x_telemetry_agent} LIKE "%Kotlin%" OR ${events.metadata__header__x_telemetry_agent} LIKE "%Swift%" OR ${events.metadata__header__x_telemetry_agent} LIKE "%Rust%")) ;;
 }
--- a/mozilla_vpn/explores/events.explore.lkml
+++ b/mozilla_vpn/explores/events.explore.lkml
@@ -1,12 +1,11 @@
 include: "//looker-hub/mozilla_vpn/explores/events.explore.lkml"
 
-explore: event_counts {
+explore: +event_counts {
   sql_always_where:  ${events.submission_date} >= '2010-01-01'
-      -- For version below 2.16 Kotlin and JS were the main Glean integrations
-      (CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_OFFSET(1)] AS INTEGER) < 2 OR (CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_OFFSET(1)] AS INTEGER) = 2 AND CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_OFFSET(2)] AS INTEGER) < 16 )
-        AND (${events.metadata__header__x_telemetry_agent} LIKE "%Kotlin%" OR ${events.metadata__header__x_telemetry_agent} LIKE "%JS%"))
-      OR
-      -- For version above 2.16 Kotlin, Swift and Rust are the main Glean integrations
-      (CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_OFFSET(1)] AS INTEGER) > 2 OR (CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_OFFSET(1)] AS INTEGER) = 2 AND CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_OFFSET(2)] AS INTEGER) >= 16)
-        AND (${events.metadata__header__x_telemetry_agent} LIKE "%Kotlin%" OR ${events.metadata__header__x_telemetry_agent} LIKE "%Swift%" OR ${events.metadata__header__x_telemetry_agent} LIKE "%Rust%")) ;;
+      AND (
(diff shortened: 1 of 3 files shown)


sean-rose (approved):
I'm approving this because I trust you know what you're doing in terms of which telemetry you want included in analyses, but it seems like doing this will come with the serious caveat that it excludes all VPN clients that haven't upgraded to a newer version using the Glean Rust/Kotlin/Swift SDKs.

In particular, it seems like significantly more VPN telemetry on iOS is still coming from Glean.js than from the Glean Swift SDK.

We could conceivably come up with more elaborate conditions to include Glean Rust/Kotlin/Swift pings starting from whichever VPN version they were added and Glean.js pings for VPN versions prior to that, but I'm not sure if it'd necessarily be worth the effort, particularly given the 6-month limit pings are kept for.

sean-rose:
cc @YeonjooSmith, @ksiegler1

brizental:
Thanks @sean-rose I think getting @YeonjooSmith 's opinion here is a great call. I realize we will miss data from previous clients. However, the latest validation I have done showed that Glean.js data coming in is not comparable to Glean.rs data -- Glean.js is underreporting significantly. I believe the best approach here would be to start new, but there would be no harm waiting a bit longer until most clients are update either.

YeonjooSmith:
@brizental I loved your investigation work and I learned so much from it. 
I see that this PR is on changing what to include events sent by Glean SDK built using `Rust` and `Swift` and exclude the ones sent by Glean.js, correct? 

I am not very familiar with what each Glean codes written by different languages do. Thanks to @sean-rose  for some background during our sync and @brizental 's investigation docs, and Sean's comment above, I gathered followings: 

* According to [Bea's investigation](https://docs.google.com/document/d/1l88NQv5V-yMgfvJGfDBMd74XGiFNYm99aO15HesxP-4/edit?usp=sharing), `Glean.rs` was designed to replace Glean.js but it turned out Glean.rs sends events that `Glean.js` was not able to. 

* `Kotlin` is used for Android and `Swift` is used for iOS(may also be used for Mac?) and they are replacing `Glean.js` from the newer version of the VPN app. 

* This PR is about to remove the data sent by `Glean.js` and include the data sent by `Swift` and `Glean.rs`. 

I think I can't form my opinion without answering the following questions. (Feel free to say my questions do not make sense or correct my assumptions): 

* Would you mind explaining why the filter includes `Swift` when `Glean.rs` generates all the data that `Glean.js`  does and more? 

* According to Sean's comment above, it seems to me that `Swift` is replacing the `Glean.js` work for iOS(and maybe Mac). Should we need to do some investigation like comparing `Glean.rs` VS `Glean.js` for `Swift`?  

* Is there a clear line on which app version that is ready swap the data from `Swift` with the data from `Glean.js` ?  If we were to do this change knowing we are losing some data, can we describe the impact and we would communicate with Santiago. (I am not notifying him here as I would like to understand the impact before communicating with him).

brizental:
Sorry for the delayed response here @YeonjooSmith. I'll go point by point.

> Would you mind explaining why the filter includes Swift when Glean.rs generates all the data that Glean.js does and more?

On mobile platforms, the part of the code that actually uploads the ping is in Swift. Even though Glean Rust may be collecting part of the metrics, all of them are tagged as Glean Swift metrics because of this implementation detail. Same for Android metrics, they are all tagged as Glean Kotlin.

> According to Sean's comment above, it seems to me that Swift is replacing the Glean.js work for iOS(and maybe Mac). Should we need to do some investigation like comparing Glean.rs VS Glean.js for Swift?

Glean.js has been turned off for iOS platforms starting on 2.15, because of it's notorious unreliability in mobile platforms. We can't do the same type of investigation because we don't have the duplicated data collection.

> Is there a clear line on which app version that is ready swap the data from Swift with the data from Glean.js ? If we were to do this change knowing we are losing some data, can we describe the impact and we would communicate with Santiago. (I am not notifying him here as I would like to understand the impact before communicating with him).

Glean Swift collection starts on 2.15. Note that we are not losing data, we would just be hiding it from the dashboards. I think it's wise to do an analysis of the new data against the old data. We know it's going to be very different and more comprehensive, but I guess verifyin these assumptions is a good idea. I have filed https://mozilla-hub.atlassian.net/browse/VPN-5134 to do that.

brizental:
We may not want to completely filter out Glean.js collection, but starting on 2.16 Glean.js is completely gone. So, we might need to do a fancier filering like "Starting from 2.16 include Glean.rs data in the views" in case we are not ready to merge this harsher filtering proposal until after 2.16 is released.

YeonjooSmith:
@brizental  Thank you for answering my questions and also filing [the ticket for the analysis](https://mozilla-hub.atlassian.net/browse/VPN-5134). I don't know how complicated it would be to set the custom filter like you mentioned: "Starting from 2.16 include Glean.rs data in the views" but  it might be worth it depending on how fast iOS users update their app.

brizental:
Ok, took a while but I am back here and I updated the queries to use Glean.rs data AFTER 2.16 and Glean.js BEFORE. Code looks kinda ugly, but it works. See screenshots below, first screenshot is without the conditional filtering and second one is with. Since 2.16. has been out for a while we are already seeing that data seems to be missing from newer pings. Notice the sharp[ drop in data after the start of August.

Screenshot 2023-08-30 at 20-58-50 Mozilla VPN Monitoring
Screenshot 2023-08-30 at 20-49-42 Mozilla VPN Monitoring

Asking for a re-review due to large code changes and would also like to know @YeonjooSmith's opinion!

mozilla_vpn/explores/events.explore.lkml
```diff
@@ -1,9 +1,12 @@
 include: "//looker-hub/mozilla_vpn/explores/events.explore.lkml"
 
-explore: +event_counts {
+explore: event_counts {
```
brizental, line 3:
```suggestion
explore: +event_counts {
```

sean-rose (changes requested):

mozilla_vpn/explores/events.explore.lkml
```diff
@@ -2,8 +2,11 @@ include: "//looker-hub/mozilla_vpn/explores/events.explore.lkml"
...
-      OR ${events.metadata__header__x_telemetry_agent} LIKE "%JS%"
-    ) ;;
+      -- For version below 2.16 Kotlin and JS were the main Glean integrations
+      (CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_OFFSET(1)] AS INTEGER) < 2 OR (CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_OFFSET(1)] AS INTEGER) = 2 AND CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_OFFSET(2)] AS INTEGER) < 16 )
+        AND (${events.metadata__header__x_telemetry_agent} LIKE "%Kotlin%" OR ${events.metadata__header__x_telemetry_agent} LIKE "%JS%"))
+      OR
+      -- For version above 2.16 Kotlin, Swift and Rust are the main Glean integrations
+      (CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_OFFSET(1)] AS INTEGER) > 2 OR (CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_OFFSET(1)] AS INTEGER) = 2 AND CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_OFFSET(2)] AS INTEGER) >= 16)
+        AND (${events.metadata__header__x_telemetry_agent} LIKE "%Kotlin%" OR ${events.metadata__header__x_telemetry_agent} LIKE "%Swift%" OR ${events.metadata__header__x_telemetry_agent} LIKE "%Rust%")) ;;
```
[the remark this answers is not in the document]
brizental: Damn, that is a sharp eye @sean-rose ! I have applied all of the review comments, code looks much nicer and works. Thanks a lot!

mozilla_vpn/explores/events.explore.lkml
```diff
@@ -2,8 +2,10 @@ include: "//looker-hub/mozilla_vpn/explores/events.explore.lkml"
 
 explore: +event_counts {
   sql_always_where:  ${events.submission_date} >= '2010-01-01'
-    AND (
-      ${events.metadata__header__x_telemetry_agent} LIKE "%Kotlin%"
-      OR ${events.metadata__header__x_telemetry_agent} LIKE "%JS%"
-    ) ;;
+      AND (
+        CASE WHEN (SAFE_CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_ORDINAL(1)] AS INTEGER) < 2 OR (SAFE_CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_ORDINAL(1)] AS INTEGER) = 2 AND SAFE_CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_ORDINAL(2)] AS INTEGER) < 16 ))
```
sean-rose, line 6: It's hard to tell what this logic is doing at a glance (especially when it's all on one line like this), so I'd suggest having some explanatory comments like you had before.
```suggestion
        -- For version below 2.16 Kotlin and JS were the main Glean integrations.
        -- For version 2.16 and above Kotlin, Swift, and Rust are the main Glean integrations.
        CASE WHEN (SAFE_CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_ORDINAL(1)] AS INTEGER) < 2 OR (SAFE_CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_ORDINAL(1)] AS INTEGER) = 2 AND SAFE_CAST(SPLIT(${events.client_info__app_display_version}, '.')[SAFE_ORDINAL(2)] AS INTEGER) < 16 ))
```
(ditto for the other explores)

sean-rose (approved):
r+wc

PR #575, opened 2023-07-05
feat(DENG-1111): replace fxa_content_* views references with fxa_all_events resources

kik-kik:
feat(DENG-1111): replace fxa_content_* views references with fxa_all_events resources

depends: https://github.com/mozilla/lookml-generator/pull/705


sean-rose (commented):
mozilla/lookml-generator#705 is also currently removing the `all_events` view, in which case `firefox_accounts/views/all_events.view.lkml` should also be removed here.

kik-kik:
We decided for now the simplest option is to just point the fxa_content_* views to fxa_all_events instead.

PR #585, opened 2023-07-24
Add App Store Data Explore

fbertsch:
Checklist for reviewer:

### DIFF
--- a/duet/duet.model.lkml
+++ b/duet/duet.model.lkml
@@ -30,3 +30,7 @@ explore: kpi_downloads {
 explore: kpi_installs{
   sql_always_where: ${period_filtered_measures} in ("this", "last");;
 }
+
+explore: app_store_territory_source_type_report {
+  label: "App Store Data Report"
+}
--- a/duet/views/app_store_territory_source_type_report.view.lkml
+++ b/duet/views/app_store_territory_source_type_report.view.lkml
@@ -1,0 +1,139 @@
+view: app_store_territory_source_type_report {
+  derived_table: {
+    sql:
+      SELECT
+        date,
+        territory,
+        source_type,
+        active_devices,
+        active_devices_last_30_days,
+        deletions,
+        installations,
+        sessions,
+        first_time_downloads,
+        redownloads,
+        total_downloads,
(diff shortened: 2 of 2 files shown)


duet/views/app_store_territory_source_type_report.view.lkml
```diff
@@ -0,0 +1,139 @@
+view: app_store_territory_source_type_report {
```
ANich, line 1: Is this the right view name? Looks like it's only the name of the first table name in the join list below

duet/duet.model.lkml
```diff
@@ -31,6 +31,10 @@ explore: kpi_installs{
   sql_always_where: ${period_filtered_measures} in ("this", "last");;
 }
 
+explore: app_store_territory_source_type_report {
+  label: "App Store Data Report"
+}
+
```
ANich, line 37: (nit: IIRC we wanted to be move these to `.explore.lkml` files)

curtismorales (approved):

duet/views/app_store_territory_source_type_report.view.lkml
```diff
@@ -0,0 +1,139 @@
...
+        page_views,
+        page_views_unique_device
+      FROM
+        mozdata.app_store.firefox_app_store_territory_source_type_report
+      FULL OUTER JOIN
+        mozdata.app_store.firefox_downloads_territory_source_type_report
+        USING (date, territory, source_type)
+      FULL OUTER JOIN
+        mozdata.app_store.firefox_usage_territory_source_type_report
+        USING (date, territory, source_type) ;;
```
ANich, line 27: There's no aggregations in the above list so this seems like it could be another derived table or a join between views in the explore no?

ANich (commented):
Left a few nits

duet/views/app_store_territory_source_type_report.view.lkml
fbertsch: I don't disagree, but I wanted it this way so the explore is more intuitive (rather than having a bunch of separate views)

PR #593, opened 2023-08-10
Minor changes for VPN acquisition funnel view and channel group proportions view file. 

YeonjooSmith:
No need to review Look ML file for `vpn_saasboard_active_subs.dashboard`. 
The code changes were from the automated code generation from the dashboard change which was to update the row limit from 1000 to 50 to resolve the visualization error.
image

The measure calculation update for `channel_group_proportions_table.view` file was to correct the aggregation method from average to sum. 

 ----------------------------------------------------------------------------------------
Checklist for reviewer:


mozilla_vpn/views/channel_group_proportions_table.view.lkml
```diff
@@ -12,7 +12,7 @@ view: +channel_group_proportions_table {
 
   measure: total_new_subscription_count_for_date {
     sql: ${TABLE}.total_new_subscriptions_for_date ;;
-    type: average
+    type: sum
```
sean-rose, line 15: I don't think `sum` is appropriate here, because [`total_new_subscriptions_for_date` is the sum of all `new_subscriptions` for the date](https://github.com/mozilla/bigquery-etl/blob/8fa35469ab2e4c2884318a5915f85d808de56307/sql/moz-fx-data-shared-prod/mozilla_vpn_derived/channel_group_proportions_live/view.sql#L57-L60), so the total of this sum of sums would be `new_subscriptions * new_subscriptions` (i.e. squared).

It seems like this measure is only useful when you're aggregating by date, so I'm guessing you're running into problems when trying to use this measure over a date range, maybe to calculate a percentage-of-total?

If you want a percentage-of-total with arbitrary aggregation I think you'll need to use a Looker table calculation to dynamically calculate the denominator (e.g. `sum(${new_subscription_count})`), or the pie chart visualization has it built in, as described in [Looker's docs](https://cloud.google.com/looker/docs/best-practices/how-to-calculate-percent-of-total).

YeonjooSmith:
some of the changes were already implemented in the past and the rest needs to be re-evaluated.

PR #616, opened 2023-09-15
Update newtab visits

mmccorks:
Added the sponsored tiles enabled pref. 

Added subsections for pocket interaction data (by position) and search interaction data (by engine and source).


sergiosonline (approved):
looks good!

PR #631, opened 2023-09-27
Rzhao credential management

rhythmofrain63:
Checklist for reviewer:

- [x] Ensure that the data is not available already (fully or partially) and recommend extending an existing dataset in favor of creating new ones. Data may be available in [bigquery-etl repository](https://github.com/mozilla/bigquery-etl), [looker-hub](https://github.com/mozilla/looker-hub) or in [looker-spoke-default](https://github.com/mozilla/looker-spoke-default/tree/e1315853507fc1ac6e78d252d53dc8df5f5f322b).
- [x] Avoid merging a PR that includes the logic of a [core metric](https://docs.telemetry.mozilla.org/metrics/index.html) or complex business logic. The recommendation is to implement core business logic in bigquery-etl. E.g. The [type of search](https://github.com/mozilla/bigquery-etl/blob/a3e59f90326816a2ecaaa3e9d5b57fe9552f7d70/sql/moz-fx-data-shared-prod/search_derived/mobile_search_clients_daily_v1/query.sql#L781) or the [calculation of DAU or visited URIs](https://github.com/mozilla/bigquery-etl/blob/9bca48821a8a0d40b1700cc14ecd8068d132ed06/sql/moz-fx-data-shared-prod/telemetry_derived/firefox_desktop_exact_mau28_by_dimensions_v1/query.sql).
- [x] Avoid merging code in Looker Explores/Views that implement analysis with multiple lines of code or that will be likely replicated in the future. Instead, aim for extending an existing dataset to include the required logic, and use [Looker aggregates](https://cloud.google.com/looker/docs/aggregate_awareness) to facilitate the analysis.
- [x] Avoid merging a PR with logic that requires validation and health checks. It is recommended to implement it in bigquery-etl for full test coverage and failure alerts.


android_default_browser_metrics.view.lkml
```diff
@@ -0,0 +1,77 @@
...
+          
+      
+      product_features AS
+      (SELECT
+          client_info.client_id,
+          DATE(submission_timestamp) as submission_date,
+          COALESCE(SUM(CASE WHEN metrics.boolean.metrics_default_browser THEN 1 ELSE 0 END)) AS metrics_default_browser
+          FROM `mozdata.fenix.metrics`
+      where DATE(submission_timestamp) >= '2021-01-01'
+      AND sample_id = 0
```
irrationalagent, line 20: is this sample_id filter intended? I notice you don't have it any other place in the query

irrationalagent: nevermind, i see you multiply by 100 below.

android_default_browser_metrics.view.lkml
```diff
@@ -0,0 +1,77 @@
...
+    datatype: date
+    sql: ${TABLE}.submission_date ;;
+  }
+
+  dimension: dau {
+    type: number
+    sql: ${TABLE}.dau ;;
+  }
+
+  dimension: metrics_default_browser_users {
```
irrationalagent, line 59: did these end up working better as dimensions than measures?

fenix/views/android_privacy_metrics_cat.view.lkml
```diff
@@ -0,0 +1,75 @@
...
+    datatype: date
+    sql: ${TABLE}.submission_date ;;
+  }
+
+  measure: preferences_etp_custom_cookies_selection {
+    type: string
+    sql: ${TABLE}.preferences_etp_custom_cookies_selection ;;
+  }
+
+  measure: metric_name {
```
irrationalagent, line 57: this might need to be a dimension? I'm not sure that measures can be strings

fenix/views/android_privacy_metrics_cat.view.lkml
```diff
@@ -0,0 +1,75 @@
...
+    drill_fields: [detail*]
+  }
+
+  dimension: submission_date {
+    type: date
+    datatype: date
+    sql: ${TABLE}.submission_date ;;
+  }
+
+  measure: preferences_etp_custom_cookies_selection {
```
irrationalagent, line 52: this might need to be a dimension? I'm not sure that measures can be strings

fenix/views/android_privacy_metrics_cat.view.lkml
```diff
@@ -0,0 +1,75 @@
+
+view: android_privacy_metrics_cat {
```
irrationalagent, line 2: could this be combined with the numeric features? one way to do that is to split out each possible value of (for example) "preferences_etp_custom_cookies_selection" into a different measure - the count of clients who have that value. however, this might not be great if there are lots of possible values.

i think the main thing with the way you have it right now is it might make it harder for others to find the specific measure they are looking for.

fenix/views/android_privacy_metrics_num.view.lkml
```diff
@@ -0,0 +1,77 @@
+
+view: android_privacy_metrics_num {
```
irrationalagent, line 2: see above on whether this can be combined with the categorical features

irrationalagent (approved):
have you verified that a 1% sample is enough data for all these measures? I don't recognize some of them so I'm just wondering if there will be enough data to get an accurate estimate in some cases.

on the other hand, if this is taking a long time, then maybe increasing the sample size is a bad idea.

some of these queries are looking pretty far back. the way this is setup now, these queries will be run new every day (I think if someone runs them once they will be cached for awhile). there [are ways](https://cloud.google.com/looker/docs/derived-tables#persistent_derived_tables_pdts) to make tables persist for a day, so that they are only run at most once per day. then, when the next day's data comes in, they are run again.

I'll approve this to unblock you from making your dashboards.

android_default_browser_metrics.view.lkml
```diff
@@ -0,0 +1,77 @@
...
+      group by 1,2),
+      
+      product_features_agg AS
+      (SELECT
+          submission_date,
+          --metrics_default_browser
+          100*COUNT(DISTINCT CASE WHEN metrics_default_browser > 0 THEN client_id END) AS metrics_default_browser_users,
+          100*COALESCE(SUM(metrics_default_browser), 0) AS metrics_default_browser,
+      FROM product_features
+      where submission_date >= '2021-01-01'
```
irrationalagent, line 30: this is probably why things are taking so long. looking all the way back to 2021.

PR #660, opened 2023-11-10
DENG-1932 - update airflow monitoring 

lelilia:
adding a duration dimension so I can show long running dags on the dashboard

### DIFF
--- a/monitoring/views/airflow_dag_run.view.lkml
+++ b/monitoring/views/airflow_dag_run.view.lkml
@@ -16,6 +16,18 @@ view: +airflow_dag_run {
     }
   }
 
+  dimension_group: run_time {
+    type: duration
+    sql_start:  ${start_raw} ;;
+    sql_end: ${end_raw} ;;
+    intervals: [
+      second,
+      minute,
+      hour,
+      day
+    ]
+  }
+
   measure: last_runs {
     type: string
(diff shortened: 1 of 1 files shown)


monitoring/views/airflow_dag_run.view.lkml
```diff
@@ -84,13 +96,13 @@ view: +airflow_dag_run {
 
   measure: avg_execution_time_sec {
     type: average
-    sql: datetime_diff(${end_raw}, ${start_raw}, second) ;;
+    sql: ${seconds_run_time};;
```
lucia-vargas-a, line 99: I'm just curious where `seconds_run_time` is defined as I don't see it in Looker-Hub or here, is it from [here](https://github.com/mozilla/looker-spoke-default/blob/7763f8719565d4b0040d9b676a7b9c3d0bdccca1/monitoring/views/airflow_dag_run.view.lkml#L24C7-L24C13)?

lelilia: yes it's coming from the group ... thought it would be nicer to group it so that I have all the options and not clutter the explore
does show the correct thing in dev so I think I did it right :)

scholtzan (approved):

monitoring/views/airflow_dag_run.view.lkml
lucia-vargas-a: Thanks, LGTM too!

lucia-vargas-a (approved):

PR #680, opened 2023-12-10
ctd downstream impact attribution

soGaussian:
This PR is to quick put data in looker so stakeholders can use the explore to visualize the data. The views uses derived sql, this is because this is simply for ad-hoc analysis and views will be deleted after delivery of report

Checklist for reviewer:

### DIFF
--- a/duet/duet.model.lkml
+++ b/duet/duet.model.lkml
@@ -22,6 +22,10 @@ explore: releases {
   }
 }
 
+explore: ctd_cohort_analysis_mobile {}
+
+explore: ctd_cohort_analysis_desktop {}
+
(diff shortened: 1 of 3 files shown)


kik-kik (approved):
Since you mention this is a one off analysis, I will approve. However, we should have a discussion on what the right process for doing these should be. One of our current issues is that we have many "one off analysis" datasets, and dashboards just laying around and never being removed. This ends up creating unnecessary clutter and making it a lot more difficult to find things.

Is there already a ticket for removing these? If not, could you please create one and link this PR to it and include a date by when this should be removed.

Thanks!

PR #683, opened 2023-12-13
Add view for measuring multi-service FxA customers

ksiegler1:
Checklist for reviewer:


akkomar (commented):
Can we build this on Glean telemetry (`accounts_frontend`/`accounts_backend`) instead of `firefox_accounts.fxa_all_events`?

ksiegler1:
I initially did this on `accounts_backend`, but the service field may be null for a lot of our new activity events because the results from our legacy events don't translate. I figured we could still use this as a historical view and de-commission once we debug what's happening on the Glean side.

akkomar (approved):

firefox_accounts/explores/fxa_multi_service_dau.explore.lkml
```diff
@@ -0,0 +1,11 @@
+include: "../views/fxa_multi_service_dau.view.lkml"
+
+explore: fxa_multi_service_dau {
+  label: "Multi-Service Active FxA"
+
+  conditionally_filter: {
+    filters: [date_date: "30 days ago for 30 days"]
+    unless: [date_week, date_month, date_quarter, date_year]
+  }
```
sean-rose, line 9: Making the filter conditional is a nice touch.

firefox_accounts/views/fxa_multi_service_dau.view.lkml
```diff
@@ -0,0 +1,58 @@
...
+
+    sql:
+    SELECT
+      DATE(timestamp) as submission_date,
+      user_id,
+      ARRAY_AGG(DISTINCT service IGNORE NULLS ORDER BY service) as services,
+      MAX_BY(country_code, timestamp) as country_code
+    FROM
+      `mozdata.firefox_accounts.fxa_all_events`
+    WHERE service IN ('sync', 'mdn-plus', 'guardian-vpn', 'fx-monitor', 'fx-private-relay', 'pocket-web', 'amo-web', 'thunderbird-addons', 'mozilla-iam', 'moz-social', 'mozilla-support', 'pontoon', 'mozilla-hubs-dev')
```
sean-rose, line 13: This would be more readable and maintainable if the services were listed in alphabetical order, or even better on separate lines in alphabetical order.

firefox_accounts/views/fxa_multi_service_dau.view.lkml
```diff
@@ -0,0 +1,58 @@
...
+    AND event_type like 'fxa_activity%'
```
sean-rose, line 14: It probably wouldn't make any practical difference in this case, but in `LIKE` expressions if you don't want an underscore to treated as a wildcard character it needs to be escaped with backslashes (two backslashes in a normal string, or one backslash in a raw string).

firefox_accounts/views/fxa_multi_service_dau.view.lkml
```diff
@@ -0,0 +1,58 @@
...
+
+  dimension: services_quantity {
+    type: number
+    sql:  ARRAY_LENGTH(${TABLE}.services) ;;
+  }
+
+  dimension: service_list {
+    type: string
+    sql:  ARRAY_TO_STRING(${TABLE}.services, ' + ') ;;
+  }
```
sean-rose, line 52: This `services_list` dimension is currently the only way to filter for particular services, using a "contains" condition.

If you want to allow more precise filtering on specific services you could define an additional view in this file like `fxa_multi_service_dau__services` that has a `service` dimension with `sql: ${TABLE} ;;`, then join that in the explore with `sql_table_name: UNNEST(fxa_multi_service_dau.services) ;;`.

sean-rose (commented):
Apologies for the belated review.

PR #703, opened 2023-12-21
DENG-2224 Telemetry Dev Cycle Dashoard

lelilia:
[ticket](https://mozilla-hub.atlassian.net/browse/DENG-2224)

don't merge before the tables are in looker-hub


telemetry_dev_cycle/explores/experiments_stats.explore.lkml
kik-kik, line 1: If we're just hiding it, should we create it to begin with?

kik-kik (approved):

telemetry_dev_cycle/explores/experiments_stats.explore.lkml
lelilia: yes because I need it for the dashboard but I don't want it to show up for people looking through the explores .. most monitoring explores are set to hidden ... if you know you can still access them, but you don't overcrowed the ui

PR #706, opened 2023-12-21
1866834 - Extending fenix_and_firefox_use_counters view (instead of creating a new one)

kwindau:
- Extended view instead of creating a new one
- Updated data type of metric to be average so the explore works better for summarizing by different dimensions


curtismorales (approved):
Looks good!

PR #724, opened 2024-01-18
increase CI timout to 15 min

lelilia:
Increasing the timeout without output for running the sync_dashboards step because it currently routinley times out

Looking back through the CI logs this routinley failed in this step for as long as the records go back. Here is a [link](https://app.circleci.com/pipelines/github/mozilla/looker-spoke-default/3363/workflows/125fe765-f6f0-48f8-bcd0-5cd7b57bba7d/jobs/1002) to a CI run from August where it timed out

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,7 @@ jobs:
       - run:
           command: |
             python ./src/sync_dashboards/main.py sync --config lookml_dashboards.yaml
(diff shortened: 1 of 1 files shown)


badboy (approved):
👍 if that fixes the issues for now.
A followup bug to investigate why this is taking so long and maybe adding some additional output so the task doesn't look stuck for so long might be a good idea.

PR #727, opened 2024-01-23
fix #725 currently still importing methods and models for sdk 31

lelilia:
when looking at this locally (I did not get super far without looker credentials) it looks like one needs to import the methods and models specifically for api40 otherswise it will still try to get them for API 3.1
image


lelilia:
https://github.com/mozilla/lookml-generator/pull/749
if this is not getting deprecated, check out this pr for hints on how to change it

kwindau (approved):

PR #755, opened 2024-02-26
Create Explore for Multi-Service Mozilla Accounts users using Glean activity events

ksiegler1:
This is a replacement for the existing Multi-Service FxA Explore ([PR](https://github.com/mozilla/looker-spoke-default/pull/683)).
It includes some changes as suggested by @sean-rose in the original PR. The numbers will look different due to the inclusion of Firefox for Android and Firefox iOS.

Checklist for reviewer:


accounts_backend/explores/mozilla_accounts_multi_service_dau.explore.lkml
```diff
@@ -0,0 +1,18 @@
+include: "../views/mozilla_accounts_multi_service_dau.view.lkml"
+
+explore: mozilla_accounts_multi_service_dau {
+  label: "Multi-Service Active Mozilla Accounts"
+
+  conditionally_filter: {
+    filters: [date_date: "30 days ago for 30 days"]
+    unless: [date_week, date_month, date_quarter, date_year]
```
sean-rose, line 8: I just recently learned that if you conditionally filter one of the dimensions from a time dimension group then Looker automatically treats the other dimensions from that time dimension group as if they're specified in `unless` ([docs](https://cloud.google.com/looker/docs/reference/param-explore-conditionally-filter#conditionally_filter_with_a_dimension_of_type_time_in_a_group_puts_the_groups_other_dimensions_in_the_unless_subparameter)), so you could remove the `unless` here and it would work the same.

That said, I'm also fine with you keeping the `unless` as is since it makes explicit what's happening.

accounts_backend/views/mozilla_accounts_multi_service_dau.view.lkml
```diff
@@ -0,0 +1,80 @@
+view: mozilla_accounts_multi_service_dau {
+
+  derived_table: {
+
+    sql:
+    SELECT
+      DATE(e.submission_timestamp) as submission_date,
+      e.metrics.string.account_user_id_sha256 AS user_id,
```
sean-rose, line 8: - Suggest keeping some indication in the column name and dimension name that this is a hashed value.
- Recommend adding a `WHERE` clause filter to only include records with non-null user ID hash values.

accounts_backend/views/mozilla_accounts_multi_service_dau.view.lkml
```diff
@@ -0,0 +1,80 @@
...
+        'MDN Plus',
+        'Mozilla VPN',
+        'Mozilla Monitor',
+        'Mozilla IAM',
+        'Mozilla Social',
+        'Mozilla Support',
+        'Mozilla Hubs',
+        'Pontoon',
+        'Pocket',
+        'Thunderbird Add-ons'
```
sean-rose, line 31: - Probably want `Add-ons` instead of `AMO Stage`.
- Maybe also `MDN Plus App`?
- It'd be nice if these were in alphabetical order (it's close but not quite).

sean-rose (approved):
r+wc

accounts_backend/views/mozilla_accounts_multi_service_dau.view.lkml
sean-rose: Also, I'd worry about how this condition will be maintained over time, as I think it'd be easy for us to miss when new services appear that should be added here or remember that this specific LookML needs to be updated.

A less precise approach of not filtering on specific service names would avoid that problem, at the cost of skewing the multi-service user counts (looking at the last week of data, not filtering on specific services would increase the multi-service user counts by ~2.3%).

ksiegler1: I was worried about overwhelming the graph with permutations of naming combinations, but I forgot I sorta by-passed this in the first place by adding a filter on the graph for having e.g user counts > 500, so the filtering is not really necessary.

PR #758, opened 2024-02-29
fix suggestions for enrollments explore

irrationalagent:
This removes the "suggest" parameters from the `slug` filter-only field from [this explore](https://mozilla.cloud.looker.com/explore/experimentation/analysis_enrollments?qid=RiCUoUY1yJOt0P9LheZcru&toggle=fil). The suggestions don't work, so it prevents you from trying to add a dashboard-level filter on this field.

### DIFF
--- a/experimentation/views/analysis_enrollments.view.lkml
+++ b/experimentation/views/analysis_enrollments.view.lkml
@@ -1,6 +1,7 @@
 view: analysis_enrollments {
   parameter: slug {
-    type: unquoted
+    type: string
+    suggest_dimension: slug_name
   }
 
   parameter: segment {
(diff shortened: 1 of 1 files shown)


irrationalagent:
this needs to be worked on still, there needs to be suggestions otherwise dashboard filters wont work.

irrationalagent:
OK, I think I was able to enable the suggestions again with a bit of hackery. There may be a better way to do this but it does what I need and shouldn't break the explore for anyone else (I hope)

experimentation/views/analysis_enrollments.view.lkml
```diff
@@ -1,10 +1,7 @@
 view: analysis_enrollments {
   parameter: slug {
-    type: unquoted
-    suggest_explore: experimenter_experiments
```
irrationalagent, line 4: What's weird about this is that if you do this:

```
suggest_explore: suggest_explore: experimenter_experiments
suggest_dimension: normandy_slug
```

The suggestions (and everything else) seems to work, but the validation fails, saying it can't find the dimension (even though it exists in the experimenter_experiments explore).

The docs do say this: 

> [in addition, ensure that the view is joined to the Explore that is specified in the suggest_explore parameter.](https://cloud.google.com/looker/docs/reference/param-field-suggest-explore)

Thus the full outer join below, even though it doesn't seem to be necessary 🤷

bochocki (approved):
lgtm

PR #764, opened 2024-03-07
feat(DENG-2913) :Updated Firefox iOS Client view and added an explore

kik-kik:
# feat(DENG-2913) :Updated Firefox iOS Client view and added an explore


fbertsch (approved):
LGTM

PR #875, opened 2024-06-10
Country level browser metrics views

bochocki:
This view is required to power an "Active Users by Country" dashboard ([draft dashboard here](https://mozilla.cloud.looker.com/dashboards/1784?Date+for+YoY+Comparison=2024%2F05%2F15)). Several complications make it infeasible to calculate without a derived table:
- It needs to use the DAU 28-Day Moving Average
- It needs query from two 28-day periods: one for the target date and one for the period 365 days prior.
- It needs to aggregate at several levels (worldwide, by region, and by country).

Checklist for reviewer:


kpi/views/country_level_yoy_metrics.view.lkml
```diff
@@ -0,0 +1,321 @@
...
+          "DAU" AS metric,
+          dau AS value,
+          LAG(dau, 1) OVER (PARTITION BY region_name, country_name, app_name ORDER BY submission_date ASC) AS prev_value,
+          dau - LAG(dau, 1) OVER (PARTITION BY region_name, country_name, app_name ORDER BY submission_date ASC) AS diff,
+        FROM
+          aggregated
+        WHERE
+          submission_date = max_date
+
+        UNION ALL
```
bochocki, line 70: All of these UNIONS are because the dashboard uses a pivot on "metric" (see below); the unions enable that format.
image

kpi/views/country_level_yoy_metrics.view.lkml
```diff
@@ -0,0 +1,321 @@
...
+          ALL ),
+      aggregated AS (
+        SELECT
+          submission_year,
+          submission_date,
+          region_name,
+          country_name,
+          app_name,
+        IF
+          (dau = 0, 1, dau) AS dau,
```
bani, line 41: why do we need this?

kpi/views/country_level_yoy_metrics.view.lkml
```diff
@@ -0,0 +1,321 @@
...
+        IF
+          (wau = 0, 1, wau) AS wau,
+        IF
+          (mau = 0, 1, mau) AS mau,
+          AVG(SUM(
+            IF
+              (dau = 0, 1, dau))) OVER (PARTITION BY submission_year, region_name, country_name, app_name) AS dau_28ma,
```
bani, line 48: this will break in January

kpi/views/country_level_yoy_metrics.view.lkml
bochocki: Some of the very small countries have 0 DAU/MAU/WAU some days. In this case, when we try to calculate the percent differences, we end up dividing by zero. I initially had the percent differences in this query but due to the way the aggregations work across dimensions it's actually more correct to do the percent differences as a table calculation; I can remove this and account for zero denominators in the table calculation.

kpi/views/country_level_yoy_metrics.view.lkml
bochocki: Good catch. I will remove the year partition and make sure the LAG diff calculation still works in the `calculated` subquery.

bochocki:
closing in favor of https://github.com/mozilla/looker-spoke-default/pull/876

PR #912, opened 2024-08-20
Push lookml to main-validation for running spectacles

scholtzan:
This change will have the CI replicate the `main` branch and push it to `main-validation`. This is needed to run LookML validation and content validation via [spectacles](https://docs.spectacles.dev/cli/tutorials/getting-started/). As it turns out spectacles cannot run the validation on the `main` branch and instead defaults to the development branch of the user running spectacles. This branch is usually not up-to-date. Looker generally doesn't allow selecting the production branch (which is `main` for this repo) in the development mode, so it's not possible to run validation on that branch. 

So as a workaround I created the `main-validation` branch which serves as a mirror of `main`. Write access will be restricted to the user used in CI. 

The plan is to use spectacles to run some Looker content validation via Airflow (after lookml-generator runs). The content validation can be set on a per folder level, so it finishes relatively quickly and validates dashboards and related explores. Hopefully, this way we can detect breakages faster.

Checklist for reviewer:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,83 +37,29 @@ jobs:
           fingerprints:
             - "SHA256:rVrow/3DEpsY1OTWFfA4PymfidEtdpLN37b8zb6lDQU"
       - run:
-          name: Early return if this is running on a branch
-          command: |
-            if [[ "$CIRCLE_BRANCH" != "main" ]]; then
-              echo "Don't push changes to main-validation as this is a PR" \
-                "so marking this step successful"
-              circleci-agent step halt
-            fi
-      - run:
           name: Push lookml to main-validation branch
           command: |
             ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -2,6 +2,7 @@
 version: 2.1
 orbs:
   python: circleci/python@1.4.0
+  docker: circleci/docker@2.2
```
BenWu, line 5: Is there something that specifically needs `docker/machine`?  Seems like it would work with any executor.

Also this version is a couple years old so it might be better to use 2.6

BenWu (approved):
r+wc

.circleci/config.yml
scholtzan: `ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts` didn't work with the `python/default` executor, so I copied the one we are using in bigquery-etl

PR #917, opened 2024-08-26
Add lookml validation to detect view wildcards

scholtzan:
Checklist for reviewer:


firefox_desktop/firefox_desktop.model.lkml
```diff
@@ -1,6 +1,125 @@
 connection: "telemetry"
 
-include: "//looker-hub/firefox_desktop/views/*.view.lkml"
+include: "//looker-hub/firefox_desktop/views/active_users_aggregates_table.view.lkml"
+include: "//looker-hub/firefox_desktop/views/adclick_history_table.view.lkml"
```
scholtzan, line 4: Adding all includes here for now. They are not all needed and should be trimmed down, but as part of this PR listing them all is identical to `include: "//looker-hub/firefox_desktop/views/*.view.lkml"` and ensures that no errors get introduced

.circleci/config.yml
```diff
@@ -31,9 +31,27 @@ jobs:
           command: |
             python ./src/sync_dashboards/main.py sync --config lookml_dashboards.yaml
           no_output_timeout: 15m
+  validate-lookml:
+    docker:
+      - image: circleci/node
```
sean-rose, line 36: This is a legacy image which is no longer being updated (the latest version has Node v17, which is several years old).  CircleCI [advises](https://circleci.com/docs/circleci-images/) using their next-generation `cimg/` images.

Also, I'd recommend pinning to avoid unexpected behavior changes.

```suggestion
      - image: cimg/node@22.7
```

.circleci/config.yml
```diff
@@ -31,9 +31,27 @@ jobs:
...
+    steps:
+      - checkout
+      - run:
+          name: Execute linting
+          command: |
+            npm set prefix ~/.npm
+            PATH="$HOME/.npm/bin:$PATH"
+            PATH="./node_modules/.bin:$PATH"
+            source ~/.bashrc
+            npm install -g @looker/look-at-me-sideways;
```
sean-rose, line 46: The next-generation `cimg/node` image appears to allow global npm package installs, so when using that image messing with the npm prefix and path shouldn't be necessary.

I'd also recommend pinning the `look-at-me-sideways` package to at least the major version to avoid unexpected behavior changes.

```suggestion
            npm install -g @looker/look-at-me-sideways@3
```

.circleci/config.yml
```diff
@@ -31,9 +31,27 @@ jobs:
...
+            if lams --reporting=no;
+              then echo "No linting errors found.";
+              else echo "Linting errors found."; exit 1;
+            fi;
```
sean-rose, line 50: I believe the trailing semicolons can be omitted.
```suggestion
            if lams --reporting=no
              then echo "No linting errors found."
              else echo "Linting errors found."; exit 1
            fi
```

.circleci/config.yml
```diff
@@ -31,9 +31,27 @@ jobs:
...
   push-lookml-for-validation:
     executor: docker/machine
     steps:
+      - checkout
```
sean-rose, line 54: Why is a checkout step being added here?  It seems unrelated, and unnecessary given how the `run` step does a manual Git glone.

sean-rose (approved):
r+wc

manifest.lkml
```diff
@@ -3,3 +3,10 @@ project_name: "spoke-default"
 local_dependency: {
   project: "looker-hub"
 }
+
+# LAMS
+# rule: no_view_wildcards {
+#  description: "Please don't use wildcard view include (e.g. include: 'views/*') to avoid increasing LookML verification time. Explicitly list views to be included."
+#  match: "$.file.*..include.*"
+#  expr_rule:  ($not ($boolean ($match "views/.*\\*" ::match))) ;;
```
sean-rose, line 11: Annoyingly it turns out that in the JSON-ified project data LAMS is matching against, if a file contains a single include then the value of `include` will be a string (not a list), which the JSONPath expression `include.*` won't match.

Here's a modification to the rule to handle `include` potentially being a string or list, which I've successfully tested locally (it finds one such single-wildcard-include case in `duet/mr1DEVELOPMENT/explores.explore.lkml`):
```suggestion
#  match: "$.file.*..include"
#  expr_rule:
#   ($let includes ($if ($is-list match) match ($list match)))
#   ($not ($any ...($map includes (-> (include) ($boolean ($match "views/.*\\*" include))))))
#  ;;
```

sean-rose (changes requested):

firefox_desktop/firefox_desktop.model.lkml
sean-rose: I think it'd be good to leave `# TODO` comments in these cases indicating that we should pare these view includes down to the minimum necessary, as the large number of them may be contributing toward longer LookML validation times.

manifest.lkml
```diff
@@ -3,3 +3,10 @@ project_name: "spoke-default"
...
+# }
```
sean-rose, line 12: IMO using a 2-space indent in this LookML-ish code like we do in most LookML would help with readability.
```suggestion
# rule: no_view_wildcards {
#   description: "Please don't use wildcard view include (e.g. include: 'views/*') to avoid increasing LookML verification time. Explicitly list views to be included."
#   match: "$.file.*..include.*"
#   expr_rule:  ($not ($boolean ($match "views/.*\\*" ::match))) ;;
# }
```

manifest.lkml
sean-rose: Or if you go with a 2-space indent like I suggested in another comment:
```suggestion
#   match: "$.file.*..include"
#   expr_rule:
#     ($let includes ($if ($is-list match) match ($list match)))
#     ($not ($any ...($map includes (-> (include) ($boolean ($match "views/.*\\*" include))))))
#   ;;
```

.circleci/config.yml
```diff
@@ -31,9 +31,23 @@ jobs:
           command: |
             python ./src/sync_dashboards/main.py sync --config lookml_dashboards.yaml
           no_output_timeout: 15m
+  validate-lookml:
+    docker:
+      - image: cimg/node@22.7
```
scholtzan, line 36:
```suggestion
      - image: cimg/node:22.7.0
```

duet/mr1DEVELOPMENT/explores.explore.lkml
```diff
@@ -1,4 +1,20 @@
-include: "views/*.view.lkml"
+include: "views/dau_newprofile.view.lkml"
```
sean-rose, line 1:
```suggestion
# TODO: remove unnecessary view imports
include: "views/dau_newprofile.view.lkml"
```

sean-rose (approved):
r+wc

PR #980, opened 2024-11-27
Creates View and Explore for feature usage events with dims

m-d-bowerman:
The existing view for `fenix.feature_usage_events` aggregates the table to submission date to join to the DAU table. This view and explore include dimensions to count feature usage events, especially by Distribution ID.


fenix/views/fenix_feature_usage_events_dims.view.lkml
```diff
@@ -0,0 +1,117 @@
+include: "//looker-hub/fenix/views/feature_usage_events.view.lkml"
+
+view: fenix_feature_usage_events_dims {
```
kik-kik, line 3: Why make a new view rather than extending the existing one we are including at the top of the file `include: "//looker-hub/fenix/views/feature_usage_events.view.lkml"`?

kik-kik: Looks like that's what you're doing now. You can do this without having to rename the view, here's an example: https://github.com/mozilla/looker-spoke-default/blob/ec0493be1efe6ca92983f85aa7ec14b3aa8c94c2/relay/views/subscriptions.view.lkml#L2

Then in this view you only need to define the new stuff and not repeat definitions for thing like `ping` or `submission`

m-d-bowerman: I renamed it intentionally, to avoid confusion with the existing view that Ruoxi's using that doesn't include any dimensions.

I was also hoping to hide the dimensions that are included in the `looker-hub` generated view, but I need to figure out how to hide dimensions by default and only show the dimensions I'm interested in. That long list of dimensions that shouldn't be used bothers me

kik-kik: I think we should just use the same table view. I don't see a good reason not to.

Regarding hidding fields, we can hide them in the view definition. Here's an example: [fenix_feature_usage_events_dims](https://github.com/mozilla/looker-spoke-default/blob/0641298c34492b0e12f7e3ec6d6f29168ab1cace/firefox_accounts/views/growth_accounting.view.lkml#L19)

Alternatively, it looks like you can hide all by default: https://github.com/mozilla/looker-spoke-default/blob/0641298c34492b0e12f7e3ec6d6f29168ab1cace/mozilla_vpn/views/mobile_subscriptions.view.lkml#L6

I think you can also hide them inside an explore: https://github.com/mozilla/looker-spoke-default/blob/0641298c34492b0e12f7e3ec6d6f29168ab1cace/firefox_accounts/explores/daily_service_users.explore.lkml#L7 (but you will have to double check this)

fenix/views/feature_usage_events.view.lkml
```diff
@@ -0,0 +1,101 @@
+include: "//looker-hub/fenix/views/feature_usage_events.view.lkml"
+
+view: +feature_usage_events {
+  label: "Firefox Android Feature Usage with Dimensions"
```
kik-kik, line 4:
```suggestion
  label: "Firefox Android Feature Usage"
```

PR #984, opened 2024-12-03
feat: Turn active_users_aggregates Looker view into a derived_table to enable parameterisation

kik-kik:
We should waiting for merging of this change until after 15th December due to ongoing KPI evaluations.

---

# feat: Turn active_users_aggregates Looker view into a derived_table to enable parameterisation

Unfortunately, it appears this is the only way for us to enable a simple drop down filter on the distribution for the end user.

### DIFF
--- a/combined_browser_metrics/views/active_users_aggregates.view.lkml
+++ b/combined_browser_metrics/views/active_users_aggregates.view.lkml
@@ -1,6 +1,47 @@
-include: "//looker-hub/combined_browser_metrics/views/active_users_aggregates.view.lkml"
 
-view: +active_users_aggregates {
+view: active_users_aggregates {
+  derived_table: {
+    sql:
+      SELECT
+        segment,
+        attribution_medium,
+        attribution_source,
+        attributed,
+        adjust_network,
+        install_source,
+        city,
+        country,
+        distribution_id,
+        first_seen_year,
+        is_default_browser,
+        channel,
+        os,
+        os_version,
+        os_version_major,
+        os_version_minor,
+        submission_date,
+        locale,
+        dau,
+        wau,
+        mau,
+        daily_users,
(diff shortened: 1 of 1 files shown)


bochocki:
Is it safe to make this changes to the existing view? Is it worth considering adding a new `derived_table` so that we don't unintentionally impact existing Looker artifacts?

lucia-vargas-a:
Looker docs say that it [doesn't automatically regenerate derived tables](https://cloud.google.com/looker/docs/reference/param-view-persist-for-for-derived-table) and would use cache (for ex. if no persistent is defined like in this case) which would not show real-time results (which unfortunately has been problematic in the past), To remove the risk, did you consider building a separate view for this derived table or using a PDT?

kik-kik:
> Is it safe to make this changes to the existing view? Is it worth considering adding a new `derived_table` so that we don't unintentionally impact existing Looker artifacts?

@bochocki so generally this should recreate what already exists. So nothing should break. The idea behind recreating this resource is that the changes automatically get applied to all existing Looker resources so no additional work is required.

There is an alternative approach that we could consider, this would include adding an additional field in the view, let's call it `distribution` for now. This would map to corresponding broader distribution, let's say `mozilla-001` could be mapped to have value of `Mozilla` in the new distribution field and then we could set up field suggestions: https://cloud.google.com/looker/docs/reference/param-field-suggestions to make sure the values we want show up in the UI.

kik-kik:
@lucia-vargas-a thanks for pointing this out. I was not aware of this Looker thing. I created a PR to pursue an alternative approach here: https://github.com/mozilla/bigquery-etl/pull/6706

kik-kik:
As per my previous comment, closing this as we decided to follow a different approach.

PR #987, opened 2024-12-06
Rename fexix_feature_usage_events to feature_usage_events to match source name + remove preaggregations.

kik-kik:
# Rename fexix_feature_usage_events to feature_usage_events to match source name + remove preaggregations.

This change will allow more flexibility to the view and allow it to be used for broader range of use cases.

### DIFF
--- a/fenix/explores/fenix_feature_usage_events.explore.lkml
+++ b/fenix/explores/fenix_feature_usage_events.explore.lkml
@@ -1,8 +1,7 @@
-include: "../views/feature_usage_events.view.lkml"
+include: "../views/fenix_feature_usage_events.view.lkml"
 include: "../views/fenix_dau.view.lkml"
 
 explore: fenix_feature_usage_events {
-  from: feature_usage_events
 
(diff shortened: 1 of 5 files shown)


fenix/explores/fenix_feature_usage_events.explore.lkml
```diff
@@ -1,15 +1,16 @@
...
+  from: feature_usage_events
 
   sql_always_where: ${fenix_dau.submission_date} >= '2022-01-01'
       AND ${fenix_feature_usage_events.submission_date} >= '2022-01-01'
       ;;
 
   join: fenix_dau {
     type: inner
-    relationship: one_to_one
+    relationship: many_to_one
```
m-d-bowerman, line 13: This join will mess up aggregations on DAU, right? The join is on submission date, so when calculating a total, it'll be a {Fenix DAU on that date} * {# of combinations of dimension values}

m-d-bowerman (approved):

PR #1081, opened 2025-06-05
betling add ctr view and explore to manufactuer_partnerships model

betling:
Checklist for reviewer:


device_manufacturer_partnerships/views/dt_ctr.view.lkml
```diff
@@ -0,0 +1,224 @@
...
+    1, 2, 3
+),
+####################
+#DT Device activations
+####################
+dt_activations AS (
+    SELECT
+        DATE(CONCAT(date, "-01")) as submission_month,
+        country_code as country,
+        preloaded as dt_preloaded,
```
m-d-bowerman, line 29: Instead of adding fields for each one's preloads, I'd add one field that records whatever for the denominator from each partnership table (e.g. SUM(preloaded) AS partner_devices for dt and SUM(activated) AS partner_devices for vivo), then include a field recording the distribution_id for the table. That'll allow you to use the filters that already exist in that dashboard to also filter the CTR tile.

m-d-bowerman (changes requested):

device_manufacturer_partnerships/views/dt_ctr.view.lkml
betling: fixed the query as we discussed.

betling:
made those changes you requested @m-d-bowerman in a new commit.  let me know if you see any other issues.

device_manufacturer_partnerships/views/dt_ctr.view.lkml
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  measure: total_ctr {
+    description: "total ctr (partner_app_open_count/partner_activated) over time"
+    type: number
+    sql: ${total_partner_app_count}/${total_partner_activated} ;;
+  }
+
+
+  # # You can specify the table name if it's different from the view name:
```
m-d-bowerman, line 135: should remove all of these comments below this line, to declutter

device_manufacturer_partnerships/views/dt_ctr.view.lkml
```diff
@@ -0,0 +1,202 @@
...
+    sql: ${TABLE}.partner_activated ;;
+  }
+
+  measure: total_partner_app_count {
+    description: "total number of clients that opened app over time"
+    type: sum
+    sql: ${TABLE}.partner_app_open_count ;;
+  }
+
+  measure: total_ctr {
```
m-d-bowerman, line 128: Did you check if this ctr will work properly in the Explore? I'm not sure if it'll properly sum and then divide or divide and then sum.

m-d-bowerman (changes requested):
One last nit, should update the name of the View and Explore to something more general (e.g. `device_partnership_app_opens` or something) to apply across all partnerships

device_manufacturer_partnerships/views/dt_ctr.view.lkml
betling: confirmed that the ctr works correctly in the explore and compared it against bigquery and get same result.  I also added some safe divides to make sure we don't divide by zero in a new commit.

m-d-bowerman (approved):

PR #1125, opened 2025-07-29

dataops-looker:


BenWu (approved):
from https://github.com/mozilla/lookml-generator/pull/1266

